PR #49, opened 2013-07-11
enable equations

jlward:


jure:
Quick question: What's the state of this work? I'm in the process of looking at a number of Docx to HTML converters and came across PyDocx.

kylegibson:
Hi Jure,

I don't believe anyone has started implementing this functionality yet.
Pull requests are definitely welcome!

Thanks,
-Kyle

PR #72, opened 2014-04-02
Not all hyperlinks are being convert correctly

jlward:
document.xml snippet
```xml
			<w:r>
				<w:fldChar w:fldCharType="begin"/>
			</w:r>
			<w:r>
				<w:instrText> HYPERLINK "https://www.google.com/"</w:instrText>
			</w:r>
			<w:r>
				<w:fldChar w:fldCharType="separate"/>
			</w:r>
			<w:r>
				<w:rPr>
					<w:rStyle w:val="style17"/>
				</w:rPr>
				<w:t>AAA</w:t>
			</w:r>
			<w:r>
				<w:fldChar w:fldCharType="end"/>
			</w:r>
```
styles.xml snippet
```xml
	<w:style w:styleId="style17" w:type="character">
		<w:name w:val="Internet Link"/>
		<w:next w:val="style17"/>
		<w:rPr>
			<w:color w:val="000080"/>
			<w:u w:val="single"/>
		</w:rPr>
	</w:style>
```
This should create the a link to google with the text AAA


kylegibson:
https://msdn.microsoft.com/en-us/library/documentformat.openxml.wordprocessing.fieldcode%28v=office.14%29.aspx

kylegibson:
http://officeopenxml.com/WPfields.php

kylegibson:
Ugh. I thought I had a good solution, and then ran into a problem. According to 17.16.18, 

> If a complex field is not closed before the end of a document story, then no field shall be generated and each individual run shall be processed as if the field characters did not exist (i.e. the contents of all field code run content shall not be displayed, and the field results shall be displayed as literal text).

This is problematic because these complex fields can span paragraphs. Meaning, if a paragraph ends in which a complex field was started but not ended, the behavior is to just transform the runs until the end of the paragraph. Unfortunately, that won't be valid if the end of the complex field never occurs.

kylegibson:
Some notes:

HYPERLINK is defined in 17.16.5.25.

Fields can be represented two ways:
* As a simple field implementation, using the fldSimple element, or
* As a complex field implementation, using a set of runs involving the fldChar and instrText elements.

This issue is implementing the second form, since that appears to be more common. The first form should not be too difficult to implement in a separate issue.

pydocx/export/base.py
```diff
@@ -94,14 +106,87 @@ def export(self):
...
         if document:
+            # process the document in two passes, since there are some cases
+            # where we can't know what to do until we look at the entire
+            # document (e.g. fields)
+            # In the first pass, discard any generated results
+            self.first_pass = True
+            for result in self.export_node(document):
+                pass
+
+            self._convert_complex_fields_into_simple_fields()
```
winhamwr: That's a pretty crummy thing for a standard to require.

pydocx/export/base.py
```diff
@@ -94,14 +106,87 @@ def export(self):
...
+                        field = wordprocessing.SimpleField(instr=field.instr, children=[])
+
+                if isinstance(child, wordprocessing.FieldChar):
+                    separate_triggered = False
+                    runs_to_remove.add(run)
+                    if child.is_type_begin():
+                        field = wordprocessing.SimpleField(children=[])
+                    elif child.is_type_end():
+                        if field is None:
+                            print('end tag without a beginning')
```
winhamwr: Do we use print statement for debug logging in other parts of pydocx? Seems like proper logging with loglevel would be nice to have.

kylegibson: We don't. I should have removed these.

pydocx/export/html.py
```diff
@@ -709,3 +711,12 @@ def export_numbering_item(self, numbering_item):
         )
         tag = HtmlTag('li')
         return tag.apply(results)
+
+    def export_field_hyperlink(self, simple_field, field_args):
+        results = self.yield_nested(simple_field.children, self.export_node)
+        if not field_args:
+            return results
```
kylegibson: Need to add a test case on this scenario

pydocx/export/base.py
```diff
@@ -367,11 +464,59 @@ def export_sdt_content_block(self, sdt_content_block):
...
+        m = re.match('^\s*([^\s]+)\s+(.*)$', instr)
+        if not m:
+            return
+        field_type = m.group(1)
+        raw_field_args = m.group(2)
+        if not raw_field_args:
+            return field_type, None
+        m = re.findall(
+            # Matches: One two "foo bar" baz
+            r'(?:\s?\s*(?:"([^"]+)"|([^\s]+))+)',
```
winhamwr: Did you get this regex from somewhere? I don't want to try and understand it if I can avoid it :)

pydocx/export/html.py
```diff
@@ -481,15 +484,14 @@ def export_deleted_text(self, deleted_text):
...
-        target_uri = hyperlink.get_target_uri()
+    def get_hyperlink_tag(self, target_uri):
         if target_uri:
             href = self.escape(target_uri)
             return HtmlTag('a', href=href)
 
     def export_hyperlink(self, hyperlink):
         results = super(PyDocXHTMLExporter, self).export_hyperlink(hyperlink)
-        tag = self.get_hyperlink_tag(hyperlink)
+        tag = self.get_hyperlink_tag(hyperlink.target_uri)
```
winhamwr: Can we call this with a named parameter here?

pydocx/export/html.py
```diff
@@ -709,3 +711,12 @@ def export_numbering_item(self, numbering_item):
...
         tag = HtmlTag('li')
         return tag.apply(results)
+
+    def export_field_hyperlink(self, simple_field, field_args):
+        results = self.yield_nested(simple_field.children, self.export_node)
+        if not field_args:
+            return results
+        # TODO this doesn't handle any switches passed into the field_args such
+        # as "\l bookmark"
+        tag = self.get_hyperlink_tag(field_args[0])
```
winhamwr: Can we call this with a named parameter here?

tests/export/html/test_field_code.py
```diff
@@ -0,0 +1,161 @@
...
+                <r>
+                    <fldChar fldCharType="begin"/>
+                </r>
+                <r>
+                    <instrText> HYPERLINK "http://www.google.com/"</instrText>
+                </r>
+            </p>
+            <p>
+                <r>
+                    <fldChar fldCharType="separate"/>
```
winhamwr: So this separates the instruction from content?

pydocx/export/html.py
winhamwr: So this is basically a malformed instrText?

winhamwr:
I can take another look at this later, but most of the code is handling the spec, which I don't fully understand. Sucks that we now have to make 2 passes.

pydocx/export/base.py
kylegibson: No, It's my own. The outer unnamed group is actually unneeded, and it can be simplified further: `\s*(?:"([^"]+)"|([^\s]+))+`

Matches all sequences that:

1. starts with 0 or more white space, followed by
2. a sequence that starts with a " and ends with a ", or
3. a sequence of one or more non-space characters

I think I want to pull this regex out into a function and put some tests on it

pydocx/export/html.py
kylegibson: That's correct.

tests/export/html/test_field_code.py
kylegibson: That's correct

pydocx/export/base.py
winhamwr: > I think I want to pull this regex out into a function and put some tests on it

That seems like a good idea. Regex bugs are easy to introduce.

winhamwr:
:+1:

PR #93, opened 2014-12-10
Process footnotes.xml

kylegibson:
Reference: ECMA 376 17.11

A basic initial implementation for footnotes.

* This doesn't process `endnotes.xml`
* `numFmt` is ignored and always uses decimals
* Footnotes are always placed at the very end of the document, regardless of `pos`. HTML has no concept of pages, but it could handle the beneath text option.
* `numRestart` and `numStart` are ignored, so the footnotes always start at 1, and never restarts

### DIFF
--- a/pydocx/tests/test_html.py
+++ b/pydocx/tests/test_html.py
@@ -9,10 +9,70 @@ from pydocx.tests import (
     WordprocessingDocumentFactory,
 )
 from pydocx.wordml import (
+    FootnotesPart,
     ImagePart,
     MainDocumentPart,
     StyleDefinitionsPart,
 )
+
+
+class FootnoteTestCase(DocumentGeneratorTestCase):
+    def test_footnote_without_definition_is_ignored(self):
+        document_xml = '''
+            <p>
+              <r>
+                <t>Foo</t>
+              </r>
+              <r>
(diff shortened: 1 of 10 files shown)


pydocx/tests/test_html.py
```diff
@@ -4,15 +4,268 @@
...
+        expected_html = '''
+            <p>
+                Foo
+                <sup>
+                    <a href="#footnote-abc" name="footnote-ref-abc">1</a>
+                </sup>
+            </p>
+            <p>Footnotes should appear below this</p>
+            <hr/>
+            <ol type="1">
```
jlward: Are we sure we want to hard code this to `type="1"`? Shouldn't we use `pydocx-decimal` or whatever the class is for decimal lists?

I don't think the semantinator will convert this correctly.

kylegibson: I did this because I thought it's what you had suggested. I can change it to whatever makes the most sense.

jlward: Ah yes, I see the comment now. I was not clear. We actually use `list-style-type` instead of `type`.

jlward:
After you update from `ol type` to `ol class` this looks good.

PR #110, opened 2015-03-06
Update PyDocX for greater usability as 3rd-party library

mfraezz:
Purpose
=======
Remove possible logging redundancy, increase flexibility in parsing as html

Changes
=======
* Remove `logging.basicConfig()`
  - This was causing duplicate werkzeug logs in the OSF
* Add property `parsed_content`
  - Having more than one `<header>` or `<footer>` tag per page isn't allowed, breaks page

Side Effects
==========
`logger.log()`'s will have to be added for continued logging.

### DIFF
--- a/pydocx/DocxParser.py
+++ b/pydocx/DocxParser.py
@@ -6,7 +6,7 @@ from __future__ import (
 )
 
 import copy
-import logging
+#import logging
 import posixpath
 
 from abc import abstractmethod, ABCMeta
@@ -39,8 +39,8 @@ from pydocx.util.xml import (
 )
 from pydocx.wordml import WordprocessingDocument
 
-logging.basicConfig(level=logging.DEBUG)
-logger = logging.getLogger("NewParser")
+#logging.basicConfig(level=logging.DEBUG)
+#logger = logging.getLogger("NewParser")
 
 
 class IterativeXmlParser(object):
--- a/pydocx/DocxParser.py
+++ b/pydocx/DocxParser.py
@@ -6,7 +6,7 @@ from __future__ import (
 )
 
 import copy
-#import logging
+import logging
 import posixpath
 
 from abc import abstractmethod, ABCMeta
@@ -39,8 +39,7 @@ from pydocx.util.xml import (
 )
 from pydocx.wordml import WordprocessingDocument
 
-#logging.basicConfig(level=logging.DEBUG)
-#logger = logging.getLogger("NewParser")
+logger = logging.getLogger("NewParser")
 
 
 class IterativeXmlParser(object):
--- a/pydocx/parsers/Docx2Html.py
+++ b/pydocx/parsers/Docx2Html.py
@@ -32,14 +32,6 @@ class Docx2Html(DocxParser):
         )
         return content
 
-    @property
-    def parsed_without_head(self):
-        content = super(Docx2Html, self).parsed
-        content = "<html><body>{body}</body></html>".format(
-            body=content,
-        )
-        return content
-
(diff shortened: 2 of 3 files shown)


pydocx/DocxParser.py
```diff
@@ -39,7 +39,6 @@
 )
 from pydocx.wordml import WordprocessingDocument
 
-logging.basicConfig(level=logging.DEBUG)
```
winhamwr: This should be moved to being called inside [main](https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/__init__.py#L22). We want to config the logger for CLI usage, still.

mfraezz: You're right, that makes sense. Made the change

pydocx/parsers/Docx2Html.py
```diff
@@ -32,6 +32,14 @@ def parsed(self):
         )
         return content
 
+    @property
+    def parsed_without_head(self):
+        content = super(Docx2Html, self).parsed
+        content = "<html><body>{body}</body></html>".format(
+            body=content,
+        )
+        return content
```
kylegibson: The preferred pattern for parser customization is to extend the class in your own code, and override/replace methods as desired. For example, you could do the following (we do something similar):

```python
class OSFDocx2HtmlParser(Docx2Html):
    def head(self):
        return ''
```

Then you would only need to do create an instance of `OSFDocx2HtmlParser` and get the `parsed` value.

How does this approach sound?

mfraezz: It sounds perfectly acceptable to me. I just thought that a fairly common recurring event in use cases for PyDocX might be displaying the parsed output in a web page, and if so an additional property to make that simpler for the end user could be worthwhile. 

I'll update this early tomorrow, after making the changes on my side of the code. Thanks for the advice :+1:

winhamwr:
It seems like there's still some room for an out of the box solution for HTML header/footer. We're also stripping them internally at PolicyStat, which is a sign that it's probably a common use case.

jlward:
I've actually been thinking about this. It would make sense to have two properties for the two use cases that I can think of. The way I see it we are either going to want to turn the docx into a fully functional webpage (which is what it does now). Or we are going to want to plug it into an existing page somewhere (we won't need the `html` or `head` tags since they will already exist on the page).

winhamwr:
Full webpage versus content within an existing page seems like a very good way to divide things.

mfraezz:
Alright, I added `parsed_content` as a property to return just the content wrapped in a `<div>`. Does this seem to you all like a good way of doing it?

I was also thinking it might be a good idea to rename `parsed` to `parsed_webpage` or something similar; though it's totally unnecessary for any reason other than clarity/readability. What are your opinions on that?

winhamwr:
Do we need to throw it in a `div`? I would see the common use-case as wanting to decide exactly where the content lives.

mfraezz:
I suppose it's not necessary to put the doc into its own container, I'll change that momentarily

mfraezz:
I don't know that the build fail is related to my change. The only error:
```

File "/home/travis/build/CenterForOpenScience/pydocx/pydocx/tests/test_xml.py", line 595, in test_performance

assert total_time < expected_time, error_message

AssertionError: Total time: 6.388738632202148; Expected time: 6
```

The test suite also consistently passes when I run it locally. Could someone retry the build?

winhamwr:
> Could someone retry the build?

Restarted. Performance tests are finicky :(

mfraezz:
Thanks. As expected, Travis just had a bad build

kylegibson:
Looks good.

PR #111, opened 2015-03-09
Rendered HTML must include utf-8 charset header

kylegibson:
The HTML rendered using `pydocx --html test.docx output.html` doesn't specify `utf-8` charset even though we output `utf-8`, so we end up with results like the following:

screenshot from 2015-03-09 12 05 19

That should instead look like:

screenshot from 2015-03-09 12 07 58


jlward:
This looks good. Way better help message for missing args for CLI.

PR #113, opened 2015-03-17
Fix bug with docx exported from google docx

arnaudlamy:
- They change the parameter pgzs from int to float


kylegibson:
Hi Arnaud,

Welcome and thank you for the pull request! Could you add a test case on this? Otherwise there's no guarantee that it won't be broken again in the future.

Thanks,
-Kyle

winhamwr:
Hello Arnaud,

Some guidance on adding tests (that should probably live on our docs):

* [running tests](https://pydocx.readthedocs.org/en/latest/development.html#running-tests)


The easiest way to add a test is to add an example .docx file that triggers this bug in [fixtures](https://github.com/CenterForOpenScience/pydocx/tree/master/pydocx/fixtures) along with a `.html` file that represents what you expect that parsed output to be.

Then, you can just add the name of that file to the [cases](https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/tests/test_docx.py#L36) attribute and it will be turned in to a testcase that parses your .docx file and compares the results against the HTML file.

The test should fail without the change in 190c5ab and then pass with it.

Thanks
-Wes

arnaudlamy:
Hi,

Thank you very much. It's done and I tested that my test crashes without my fix.

Best,
Arnaud

pydocx/tests/test_docx.py
```diff
@@ -123,6 +123,13 @@ def test_result_from_file_pointer_matches_result_from_path(self):
         assert file_html
         self.assertEqual(path_html, file_html)
 
+    def test_googledocx_parsing(self):
+        try:
+            Docx2Html(path=self.get_path_to_fixture('export_from_googledocs.docx')).parsed
+        except ValueError:
+            self.fail("Can't parse google docs exported docx")
```
kylegibson: Instead of adding this separate test case, you can add `'export_from_googledocs'` to the list here: https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/tests/test_docx.py#L36. Test cases are automatically created from that list of input/output files.

pydocx/fixtures/export_from_googledocs.html
```diff
@@ -0,0 +1 @@
+<html><head><meta charset="utf-8" /><style>.pydocx-caps {text-transform:uppercase}.pydocx-center {text-align:center}.pydocx-comment {color:blue}.pydocx-delete {color:red;text-decoration:line-through}.pydocx-hidden {visibility:hidden}.pydocx-insert {color:green}.pydocx-left {text-align:left}.pydocx-right {text-align:right}.pydocx-small-caps {font-variant:small-caps}.pydocx-strike {text-decoration:line-through}.pydocx-tab {display:inline-block;width:4em}.pydocx-underline {text-decoration:underline}body {margin:0px auto;width:51.00em}</style></head><body><p>Beauty will save the world.</p></body></html>
```
kylegibson: This should just be `<p>Beauty will save the world.</p>`, see the other fixture html files for examples.

arnaudlamy:
Sorry about that, I didnt' get it. Done

PR #112, opened 2015-03-17
Support using defusedxml if it's available

kylegibson:
See https://pypi.python.org/pypi/defusedxml/

### DIFF
--- a/pydocx/tests/__init__.py
+++ b/pydocx/tests/__init__.py
@@ -146,8 +146,6 @@ class WordprocessingDocumentFactory(object):
         StyleDefinitionsPart: 'prepare_style_content',
     }
 
-    xml_header = '<?xml version="1.0" encoding="UTF-8"?>'
-
     relationship_format = '''
         <Relationship Id="{id}" Type="{type}" Target="{target}" TargetMode="{target_mode}"/>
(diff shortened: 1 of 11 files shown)


pydocx/tests/__init__.py
```diff
@@ -160,11 +158,15 @@ class WordprocessingDocumentFactory(object):
...
     '''  # noqa
 
-    def __init__(self, items=None):
+    def __init__(self, items=None, xml_header=None):
         self.items = items
         if not self.items:
             self.items = {}
 
+        if xml_header is None:
+            xml_header = '<?xml version="1.0" encoding="UTF-8"?>'
```
michaelgzoller: If you're overriding the default, why not just make this string the default instead of None?

michaelgzoller:
Looks good to me

pydocx/tests/__init__.py
kylegibson: It's a development pattern I use. I always initialize defaults within the function rather than the signature. It keeps the function signature cleaner, and it works for every case. http://effbot.org/zone/default-values.htm

kylegibson:
Can you move the trello card to Deploy?

PR #114, opened 2015-03-19
Clean up the pydocx module by moving the tests to its own root directory

kylegibson:
For example:

https://github.com/django/django
https://github.com/PolicyStat/terrarium
https://github.com/jakubroztocil/httpie
https://github.com/mitsuhiko/flask
https://github.com/getsentry/sentry

Maybe take some time in this ticket to review http://www.jeffknupp.com/blog/2013/08/16/open-sourcing-a-python-project-the-right-way/ and either do that, or make new issues.

Even the project from the above article:

https://github.com/jeffknupp/sandman


michaelgzoller:
Looks good to me.

JeffSpies:
+1

run_tests.sh
```diff
@@ -1,3 +0,0 @@
-#! /bin/sh
-
-nosetests --verbose --with-doctest --with-coverage --cover-package pydocx $@ && find pydocx -name '*.py' | xargs flake8
```
jlward: Why did you remove this file? I really like having a one liner that runs the tests faster than tox does.

kylegibson: How to do you feel about moving this command to a `Makefile`? It's slightly more standard than this approach. You would type `make test` to run the tests. Thoughts?

jlward: I would not be opposed to that.

michaelgzoller:
Still looks good to me.

PR #105, opened 2014-12-19
release files after parsed

tanyunshi:
fix for https://github.com/CenterForOpenScience/pydocx/issues/104

PR #115, opened 2015-03-20
Move CLI to __main__.py

### LINKED ISSUE #104: DocxParser: input docx is not completely closed after parsed
The docx file is not completely closed/released  after parsed by DocxParser. As a result, we cannot delete the docx after being parsed.

For example, the following code gives exception as the file test.docx cannot be deleted as it's used by other process.

    infile = os.path.join(os.path.dirname(__file__), 'docx/test.docx')
    parser = Docx2Html(path=infile)
    print(parser.parsed)
    shutil.rmtree(os.path.join(os.path.dirname(__file__), 'docx'))

I think after the docx is parsed, we should close le ZipExtFile extracted from the zipfile in order to release the docx, for example in DocxParse.py:

    @property
    def parsed(self):
        if not self._parsed:
            self._load()
          
        document_streams = self.document.package.streams
        for ref in document_streams.keys():
             document_streams[ref].close()
       
        return self._parsed

kylegibson:
See https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/

This will resolve #104, #105, #106,


pydocx/__main__.py
```diff
@@ -0,0 +1,48 @@
...
+import logging
+
+from pydocx.parsers.Docx2Html import Docx2Html
+from pydocx.parsers.Docx2Markdown import Docx2Markdown
+
+
+def convert(parser_type, docx_path, output_path):
+    if parser_type == '--html':
+        output = Docx2Html(docx_path).parsed
+    elif parser_type == '--markdown':
```
michaelgzoller: In one of your commit messages you said Markdown support wasn't great, and skipped that test. Do you still want to keep the flag active here?

kylegibson: Yeah, I wasn't sure about what to do with that, so I figured leaving it the way it is was the way to go, for now.

pydocx/__main__.py
```diff
@@ -0,0 +1,48 @@
...
+    else:
+        print('Only valid parsers are --html and --markdown')
+        return 2
+    with open(output_path, 'wb') as f:
+        f.write(output.encode('utf-8'))
+    return 0
+
+
+def usage():
+    print('Usage: pydocx [--html|--markdown] input.docx output')
```
michaelgzoller: Conventionally, brackets indicate optional arguments. Required arguments are not inside brackets.

`Usage: pydocx --html|--markdown input.docx output`

michaelgzoller:
This looks good to me, bearing in mind my note about the usage string.

PR #116, opened 2015-03-29
Don't assume the first sz of an rPr actually is a direct child of that rPr

kylegibson:
The `get_properties_with_no_font_size` function in `parse_r` incorrectly assumes (using `find_first`) that the first `sz` element is actually a direct child of the `rPr` when it tries to remove it. This assumption is proven false because an `rPr` may contain a `rPrChange` which itself may contain a `rPr`, and that final `rPr` may have a `sz`. Consequently, the following traceback occurs. 

This is yet another issue as a result of our use of the `find_first` helper. 

```
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/parsers/Docx2Html.py", line 27, in parsed
    content = super(Docx2Html, self).parsed
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 819, in parsed
    self._load()
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 229, in _load
    self.parse_begin(main_document_part)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 264, in parse_begin
    self._parsed = self.parse(main_document_part.root_element)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 267, in parse
    return self.parser.parse(el)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 145, in parse
    el,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 105, in parse
    stack,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 153, in process_tag_completion
    result = func(element, result, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 461, in parse_p
    return self.parse_list(el, text, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 325, in parse_list
    parsed = self._parse_list(el, text, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 391, in _parse_list
    parsed += self.parse(next_el)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 267, in parse
    return self.parser.parse(el)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 145, in parse
    el,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 105, in parse
    stack,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 153, in process_tag_completion
    result = func(element, result, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 465, in parse_p
    return self.parse_list_item(el, text, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 542, in parse_list_item
    next_elements_content = self.parse(el)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 267, in parse
    return self.parser.parse(el)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 145, in parse
    el,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 105, in parse
    stack,
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 153, in process_tag_completion
    result = func(element, result, stack)
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 754, in parse_r
    paragraph_properties = get_properties_with_no_font_size()
  File "/home/policystat/env/lib/python2.6/site-packages/pydocx/DocxParser.py", line 747, in get_properties_with_no_font_size
    rpr.remove(size_tag)
ValueError: list.remove(x): x not in list
```


tests/test_docx.py
```diff
@@ -45,8 +45,6 @@ class ConvertDocxToHtmlTestCase(TestCase):
         # level upperRomans are converted.
         'external_image',
         'export_from_googledocs',
-        'fake_subscript',
```
jlward: I feel like we should not be removing these test cases. We historically have at least one test case that uses a real docx to make sure our broad assumptions about docx files are valid.

tests/test_html.py
```diff
@@ -397,6 +397,307 @@ def test_non_entity_unicode_character(self):
         self.assert_document_generates_html(document, expected_html)
 
 
+class FakedSubScriptTestCase(DocumentGeneratorTestCase):
+    style_xml = '''
+        <style styleId="foo" type="paragraph">
```
jlward: Can we use a `styleId` that could actually show up? Instead of `foo`.

jlward:
Looks good with suggestions.

tests/test_docx.py
kylegibson: My concern with the fixture approach is that docx fixtures are much more difficult to manage and audit than snippets of XML. As the number of docx fixtures increase, this problem necessarily becomes worse. 

However, I think we definitely want to ensure that if we remove fixture files, we're not losing or missing out on something. I don't think we'll ever remove all of the fixture files, but I feel that reducing or restricting their number, where appropriate, is a good path. To explain why, and hopefully mitigate your concerns, I have the following additional thoughts:

The `DocumentGeneratorTestCase` uses the provided XML to generate valid docx containers as per the specification. That doesn't guarantee that the XML itself is valid, since any arbitrary XML can be used. For that, we continue to use Word/Libreoffice to generate the XML, just like we would for generating a fixture file. However, instead of storing the fixture as a single file, we extract the relevant XML and make it into a test case. That's what I've done in this ticket. I used the relevant XML present in the fixture to form the basis of the new test cases. 

Regarding assumptions, the primary assumption we're working from is that Word produces EMCA-376 valid documents. This assumption isn't arbitrary. That doesn't mean it can't eventually be proven false. However, I don't think one of our docx fixtures will accidentally capture this.

tests/test_html.py
kylegibson: Why you think `foo` couldn't show up? Do you mean something like `'Green'` or `'Base'` or `'TestStyle'` (these examples appear in the documentation).

tests/test_docx.py
jlward: I don't think they are difficult to manage. As long as their names describe what should be testing (which for the most part is done), then all you have to do is open the file to see make sure it actually doing what we think it's doing. I am still quite against removing these files because we are testing a real life document to make sure our implementation works on a real document. Not just he XML we create (that may or may not actually be correct).

tests/test_html.py
jlward: I find it quite unlikely that MS Word or Libreoffice will ever auto-create a style called `foo`

kylegibson: I agree. However, you can create custom styles, and `foo` is an acceptable name for such a style. The name you use is also used as the `styleId`.

jlward: I would say that `foo` is not a good variable name.

kylegibson: It's not a variable name. It's a styleId, and a valid one, that libreoffice allows me to use to create a new style. Do you have a suggestion for an alternative that conveys that the styleId can literally be anything?

jlward: How about `fake_subscript`

tests/test_docx.py
kylegibson: I reverted the commits that removed the fixtures, so they are back now.

PR #117, opened 2015-03-30
Eliminate all improper usages of the find_first utility function

kylegibson:
It's usage is problematic, since it encourages invalid assumptions about XML structure. Depends on #116


jlward:
Can you merge this up with master to make reviewing easier?

pydocx/util/xml.py
```diff
@@ -40,6 +40,8 @@ def el_iter(el):
 def find_first(el, tag):
     """
     Find the first occurrence of a tag beneath the current element.
+
+    This method should not be used to navigate an OOXML schema
```
jlward: How do you feel about adding a `DepricationWarning` here?

jlward:
This looks fine.

pydocx/util/xml.py
kylegibson: We could do that... although technically since we're < 1.0 it's not required. In alpha, everything is pending deprecation.

jlward: Lol, fair enough.

PR #121, opened 2015-04-06
Update WordprocessingDocumentFactory to handle Numbering definitions

kylegibson:
Also update the DocumentGeneratorTestCase to handle additional explicit arbitrary parts

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ docs:
 	$(MAKE) -C docs html
 
 release: clean
-	python setup.py sdist bdist_wheel upload
+	python setup.py sdist upload
 
 sdist: clean
 	python setup.py sdist
--- a/tests/export/test_html.py
+++ b/tests/export/test_html.py
@@ -10,6 +10,7 @@ from pydocx.wordml import (
     FootnotesPart,
     ImagePart,
     MainDocumentPart,
+    NumberingDefinitionsPart,
     StyleDefinitionsPart,
 )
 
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -36,7 +36,7 @@ docs:
 	$(MAKE) -C docs html
 
 release: clean
-	python setup.py sdist bdist_wheel upload
```
jlward: I don't even know what `bdist_wheel` does.

kylegibson: It came with the `Makefile` template I used. It doesn't work, so I removed it for now. We can always add it back later if for some reason it was actually useful.

tests/export/test_html.py
```diff
@@ -1678,3 +1817,41 @@ def test_with_line_break(self):
 
         expected_html = '<p><a href="http://google.com">li<br />nk</a>.</p>'
         self.assert_document_generates_html(document, expected_html)
+
+
+class NumberingTestCase(DocumentGeneratorTestCase):
+    def test_lowerLetter_numbering_format_is_handled(self):
+        num_id = '2'
+        numbering_xml = '''
+            <num numId="{num_id}">
```
jlward: I assume this is was the thing you showed me the other day?

jlward:
Looks good.

tests/export/test_html.py
kylegibson: No, it's not. I added support for handling numbering definitions to the DocumentGeneratorTestCase. It didn't make sense to do that without having at least one test case in the pydocx repo that used it, so I added one.

jlward: Makes sense.

PR #119, opened 2015-04-03
Create a new "export" module

kylegibson:
* Move `pydocx.DocxParser` to `pydocx.export.base`
* Move `pydocx.parsers.Docx2Html` to `pydocx.export.html`
* Move `pydocx.parsers.Docx2Markdown` to `pydocx.export.markdown`
* Rename the `DocxParser` class to `PyDocXExporter`
* Rename the `Docx2Html` class to `PyDocXHTMLExporter`
* Rename the `Docx2Markdown` class to `PyDocXMarkdownExporter`
* Add a new module `pydocx.pydocx` with a class `PyDocX` that has the methods:
  * `to_html`
  * `to_markdown`

PR #122, opened 2015-04-06
Usage documentation is no longer valid

kylegibson:
This is a regression for #119

### DIFF
--- a/pydocx/export/html.py
+++ b/pydocx/export/html.py
@@ -338,8 +338,8 @@ class PyDocXHTMLExporter(PyDocXExporter):
 
     def _convert_measurement(self, value):
         '''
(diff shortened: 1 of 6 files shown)


pydocx/export/html.py
```diff
@@ -338,8 +338,8 @@ def page_break(self):
 
     def _convert_measurement(self, value):
         '''
-        >>> parser = PyDocXHTMLExporter('foo')
-        >>> parser._convert_measurement(30)
+        >>> exporter = PyDocXHTMLExporter('foo')
```
winhamwr: :+1:

better name.

winhamwr:
Looks good.

PR #107, opened 2015-04-13
list-style-type is not a valid DOM attribute

jlward:
`ordered_list` in `Docx2Html` sets `list-style-type` directly on the DOM element. This is not valid HTML:

```html
<ol list-style-type="decimal">
   <li></li>
</ol>
```

Instead, it should do the following:

```html
<ol style="list-style-type: decimal">
   <li></li>
</ol>
```


kylegibson:
Looks good

PR #126, opened 2015-04-20
Split up the wordml and openxml modules

kylegibson:
This is just a refactoring issue. There is no new code introduced.

``pydocx.openxml`` -> ``pydocx.openxml.packaging.*``
``pydocx.wordml`` -> ``pydocx.openxml.packaging.*``


pydocx/openxml/packaging/main_document_part.py
```diff
@@ -0,0 +1,65 @@
...
+    '''
+    Represents the actual document XML tree within a Word document container.
+    This OpenXmlPart exposes several child parts for styles, numbering, fonts
+    and images.
+
+    See also: http://msdn.microsoft.com/en-us/library/documentformat.openxml.packaging.maindocumentpart%28v=office.14%29.aspx  # noqa
+    '''
+
+    relationship_type = '/'.join([
+        'http://schemas.openxmlformats.org',
```
michaelgzoller: This link does not resolve. Does that matter?

michaelgzoller:
Looks good, @kylegibson

pydocx/openxml/packaging/main_document_part.py
kylegibson: No, but good for checking. http://en.wikipedia.org/wiki/XML_namespace#Namespace_names

XML just uses URIs for namespacing, because they are guaranteed to be globally unique. 

> In general, however, users should assume that the namespace URI is simply a name, not the address of a document on the Web.

PR #127, opened 2015-04-21
Merge StylesManager into the StyleDefinitionsPart, add styles attribute

kylegibson:
The functions defined in ``pydocx.manager.styles`` could just as well live on the ``StyleDefinitionsPart``.

Additionally, move ``pydocx.models.styles`` to ``pydocx.openxml.wordprocessing.styles`` and add a styles attribute to ``StyleDefinitionsPart``.

See https://msdn.microsoft.com/en-us/library/documentformat.openxml.packaging.styledefinitionspart_members%28v=office.14%29.aspx

PR #128, opened 2015-04-21
Add an XmlCollection type, and update Styles to use it

kylegibson:
Currently Styles has to implement a collection-like interface. Abstract this implementation into XmlModel and define a XmlCollection type so other elements can benefit from it. 

This depends on #127


pydocx/models.py
```diff
@@ -43,6 +46,36 @@ def __init__(self, name=None, default=None, type=None, attrname=None):
...
 
+class XmlCollection(XmlField):
+    '''
+    Represents a collection of elements. To define a field of this type,
+    a mapping dictionary must be passed in specifying the name of the child
+    elements and a callable handler for each.
+
+    Example:
+
+    class ParkingLot(XmlModel):
```
winhamwr: :+1:
I really like this example!

pydocx/models.py
```diff
@@ -142,14 +182,42 @@ def child_handler(child):
...
-                handler = child_handlers.get(child.tag, None)
-                if callable(handler):
-                    kwargs[field_name] = handler(child)
+        # Build a mapping of tag names to collections
+        collection_member_to_collections = defaultdict(list)
+        for field_name, field in collections.items():
+            for tag_name in field.name_to_type_map.keys():
+                collection_member_to_collections[tag_name].append(field_name)
+
+        if element:
```
winhamwr: Why do we want this conditional? `for child in element` should do this implicitly, right?

kylegibson: Not if element is `None`, which it could be.

tests/openxml/wordprocessing/test_run_properties.py
```diff
@@ -0,0 +1,73 @@
+# coding: utf-8
```
winhamwr: :+1: to `RunProperties` having their own test module.

tests/util/test_zip.py
```diff
@@ -0,0 +1,22 @@
...
+    unicode_literals,
+)
+
+from unittest import TestCase
+
+from pydocx.exceptions import MalformedDocxException
+from pydocx.util.zip import ZipFile, BytesIO
+
+
+class ZipFileTestCase(TestCase):
```
winhamwr: I was a bit surprised to see this test added in this PR.

winhamwr:
Looks good!

tests/util/test_zip.py
kylegibson: I was waiting on review, so I added a few missing test cases to help with code coverage.

PR #129, opened 2015-04-23
Implement the Numbering class for the NumberingDefinitionsPart

kylegibson:
https://msdn.microsoft.com/en-us/library/documentformat.openxml.wordprocessing.numbering%28v=office.14%29.aspx

Depends on #128


pydocx/openxml/packaging/numbering_definitions_part.py
```diff
@@ -20,3 +21,14 @@ class NumberingDefinitionsPart(OpenXmlPart):
...
+    def __init__(self, *args, **kwargs):
+        super(NumberingDefinitionsPart, self).__init__(*args, **kwargs)
+        self._numbering = None
+
+    @property
+    def numbering(self):
+        if self._numbering:
+            return self._numbering
+        self._numbering = Numbering.load(self.root_element)
+        return self._numbering
```
kylegibson: I need to add a test on this

tests/openxml/wordprocessing/test_level.py
```diff
@@ -0,0 +1,83 @@
...
+        self.assertEqual(level.paragraph_style, "normal")
+
+    def test_run_properties_child(self):
+        xml = b'''
+            <lvl>
+                <rPr />
+            </lvl>
+        '''
+        level = self._load_from_xml(xml)
+        self.assertEqual(level.run_properties.__class__, RunProperties)
```
jlward: Comparing classes in this way is not the correct way of doing it. You should be using `isinstance` instead. For all cases in this diff.

jlward:
After you update the test cases to use `isinstance`, this looks good.

tests/openxml/wordprocessing/test_level.py
kylegibson: Could you help me understand why it's not the correct way, and why `isinstance` is preferred in this particular case, and all cases in this diff?

If I was doing control flow based on the type of an object I would use `isinstance` (for example, if obj is a string, do X) but in this case I am specifically wanting to ensure the class is what it should be.

Also, the benefit of doing it this way is that when the comparison fails, the assertion shows what the actual class is. I don't get that with `isinstance`.

jlward: http://www.siafoo.net/article/56

jlward: https://www.python.org/dev/peps/pep-0008/

kylegibson: The first resource makes a solid case that `isinstance` is preferred to `__class__` *when you are writing control flow logic*. Totally agree. I'm not doing that here. 

Also, as confirmed by the first resource, `isinstance` would return true if the class is a sub-class. In the model implementation schema, I feel that would be a bug. The class should not be a sub-class (of `RunProperties` in this case), but the exact class itself.

The PEP8 resource doesn't mention `__class__` at all. Instead, it says you should not use `type` in a conditional. I'm not using type, and I don't have a conditional. It doesn't say that `__class__` is to be avoided.

Just so we're on the same page, the alternative we're discussing is the following:

```python
assert isinstance(level.run_properties, RunProperties)
```

Downsides:

* When the assertion fails, there's no visibility to what the actual class is, just what it is not
* Captures sub-classes, which may not be desired (I feel they are not)

Upsides:

* Avoids having to access a dunder property
* This is how type checking is preferred for control flow

Thoughts?

jlward: > When the assertion fails, there's no visibility to what the actual class is, just what it is not

You could do something like

```python
assert isinstnace(level.run_properties, RunProperties), level.run_properties
```

Or even make a helper method that does this for you. That way, if your __repr__ method on the classes are set up correctly, you will know what class it actually is.

>  ... and I don't have a conditional

What do you think self.assertEqual uses behind the scenes? :) (Hint, it's a conditional)

kylegibson: How would I solve the sub-class issue? `insinstance` would return true for a sub-class. Wouldn't that be a bug?

jlward: Is it a bug? I don't think it is. sub classes are should be fine.

PR #130, opened 2015-04-24
Update documentation to indicate what portions of the ECMA numbering spec are actually supported

kylegibson:
I was thinking of using a table. Each row would correspond to a particular section in the specification (denoted by the section identifier, such as `17.9.1`), followed by a blurb for the section, and then to what degree PyDocX provides support for it. Would be cool if each row could link to some external resource.

Also, it may be useful for top-level sections to display a percentage of support? So `17.9` would display the percentage support of it's sub-level sections.

### DIFF
--- a/docs/conformance.rst
+++ b/docs/conformance.rst
@@ -1,0 +1,42 @@
+###########
+Conformance
+###########
+
+Open Office XML is standardized by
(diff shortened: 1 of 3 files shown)


jlward:
This looks good.

PR #131, opened 2015-04-29
XML models should specify their own tag names, not parents

kylegibson:
Currently, the tag handled by a child is dictated by the parent. This is backwards. The child should be aware of the tag that it handles.


jlward:
This looks good.

PR #132, opened 2015-04-29
Make importing PyDocX easier

kylegibson:
Currently you have to do `from pydocx.pydocx import PyDocX` which is somewhat redundant. Make it simpler: `from pydocx import PyDocX`


jlward:
Looks good

PR #135, opened 2015-05-05
Decouple header processing from the pre-processor

kylegibson:
The header processing is very HTML-centric, but it lives in the base pre-processor. Decouple this logic from the pre-processor, and split it appropriately between the base exporter and html exporter.

### DIFF
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -267,7 +267,6 @@ class PyDocXExporter(MultiMemoizeMixin):
 
         self.pre_processor = self.pre_processor_class(
             convert_root_level_upper_roman=self.convert_root_level_upper_roman,
-            styles=self.document.main_document_part.style_definitions_part.styles,  # noqa
             numbering_root=self.numbering_root,
         )
         self.pre_processor.perform_pre_processing(main_document_part.root_element)  # noqa
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -201,12 +201,12 @@ class PyDocXExporter(MultiMemoizeMixin):
     def parse_run_properties(self, el, parsed, stack):
         properties = RunProperties.load(el)
         parent = stack[-1]['element']
-        self.document.main_document_part.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+        self.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
 
     def parse_paragraph_properties(self, el, parsed, stack):
         properties = ParagraphProperties.load(el)
         parent = stack[-1]['element']
-        self.document.main_document_part.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+        self.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
 
     @property
     def document(self):
@@ -233,18 +233,15 @@ class PyDocXExporter(MultiMemoizeMixin):
             return self.main_document_part.numbering_definitions_part
 
     def _load(self):
-        self.document = WordprocessingDocument(path=self.path)
-        main_document_part = self.document.main_document_part
-        if main_document_part is None:
+        if self.main_document_part is None:
             raise MalformedDocxException
 
         self.numbering_root = None
-        numbering_part = main_document_part.numbering_definitions_part
-        if numbering_part:
-            self.numbering_root = numbering_part.root_element
+        if self.numbering_definitions_part:
+            self.numbering_root = self.numbering_definitions_part.root_element
 
-        self.page_width = self._get_page_width(main_document_part.root_element)
-        self.parse_begin(main_document_part)
+        self.page_width = self._get_page_width(self.main_document_part.root_element)  # noqa
+        self.parse_begin(self.main_document_part)
 
     def load_footnotes(self, main_document_part):
         footnotes = {}
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -171,6 +171,8 @@ class PyDocXExporter(MultiMemoizeMixin):
         self.footnote_ordering = []
         self.current_part = None
 
+        self._document = None
+
         self.parse_tag_evaluator_mapping = {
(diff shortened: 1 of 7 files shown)


pydocx/export/base.py
```diff
@@ -199,26 +201,47 @@ def __init__(
...
 
     def parse_paragraph_properties(self, el, parsed, stack):
         properties = ParagraphProperties.load(el)
         parent = stack[-1]['element']
-        self.document.main_document_part.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+        self.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+
+    @property
+    def document(self):
+        if self._document:
```
winhamwr: I was going to make a comment about the downsides of doing an operation that might fail in a property, but I see that this is just a better-structured version of what was already happening in `_load`.

pydocx/export/base.py
```diff
@@ -779,7 +826,7 @@ def paragraph(self, text):
         return text
 
     @abstractmethod
-    def heading(self, text, heading_level):
+    def heading(self, text, heading_style_name):
```
winhamwr: This makes this a backward-incompatible change, right? Shouldn't we document that change in the CHANGELOG and call it out, explicitly? Any customer parsers will need to update their definition.

pydocx/export/html.py
```diff
@@ -22,6 +22,30 @@
...
+            'heading 2': 'h2',
+            'heading 3': 'h3',
+            'heading 4': 'h4',
+            'heading 5': 'h5',
+            'heading 6': 'h6',
+        }
+        self.default_heading_level = 'h6'
+
+    def load_document(self):
+        document = super(PyDocXHTMLExporter, self).load_document()
```
winhamwr: The base class assigns this to `self._document`. Does that matter? Looks like the `document` property might not cache the result, otherwise.

pydocx/export/base.py
```diff
@@ -199,26 +201,47 @@ def __init__(
...
         properties = ParagraphProperties.load(el)
         parent = stack[-1]['element']
-        self.document.main_document_part.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+        self.style_definitions_part.save_properties_for_element(parent, properties)  # noqa
+
+    @property
+    def document(self):
+        if self._document:
+            return self._document
+        return self.load_document()
```
winhamwr: To avoid the problem where `load_document` in a subclass forgets to assign to `self._document`, what about doing that here, also?

pydocx/export/html.py
kylegibson: Yes, it does matter.

pydocx/test/utils.py
```diff
@@ -282,8 +282,8 @@ def __init__(self, *args, **kwargs):
         super(XMLDocx2Html, self).__init__(path=None, *args, **kwargs)
 
     def _load(self):
-        self.document = WordprocessingDocument(path=None)
-        package = self.document.package
+        document = WordprocessingDocument(path=None)
```
winhamwr: Is this necessary? Can't we just do `package = self.document` on the next line?

pydocx/export/base.py
kylegibson: Do you think we should do that even though we haven't released a stable public API? We're not 1.0.0 yet.

pydocx/test/utils.py
kylegibson: Yes, unfortunately it is necessary. The reason we can't do that is because the document hasn't been built yet. Calling self.document will call load_document, which in turn queries for style_definitions_part which depends on main_document_part. This triggers the container to load/process the parts if it hasn't already. This document doesn't have any parts, so it becomes empty. Adding the stuff below doesn't cause the load parts to be invalidated so it isn't re-triggered. I didn't want to have to rewrite that now, so I worked around it.

pydocx/export/base.py
kylegibson: I'm not sure I follow what you mean, can you elaborate, or give an example?

pydocx/test/utils.py
winhamwr: Maybe a little comment to that effect would be useful, then?

kylegibson: What I think I'd like to do is replace all of this logic with a WordprocessingDocumentFactory, but that would require updating a bunch of tests since the spec for that is somewhat different than what's being used here.

pydocx/export/base.py
winhamwr: I think so. We don't need to bump the major version, since we're not 1.0.0, but it's pretty easy to communicate something like this and would save our early users (usually the most important users) a lot of potential headache.

pydocx/export/base.py
winhamwr: I'm talking about changing this to:
```python
def document(self):
    if self._document:
        return self._document
    self._document = self._load_document()
    return self._document
```

winhamwr: Edit: _load_document => load_document

kylegibson: Oh I see. So if a subclass overrides `load_document` but forgets to call super.

pydocx/export/html.py
winhamwr: So you're trying to avoid using `self._document`, here? I'm sure I'm missing something, but my reading of this right now is that every reference to `self.document` inside `PyDocXHTMLExporter` will result in a new call to `load_document`, which creates a new `WordprocessingDocument`, which sounds expensive.

winhamwr: NM. The `super()` call takes care of the `self._document` assignment.

pydocx/export/base.py
winhamwr: Right.

winhamwr:
I'd like to see a backward-incompatibility note in the CHANGELOG. Looks good, otherwise.

winhamwr:
:+1:

PR #136, opened 2015-05-06
Convert the 'Root level upper roman list to header' handling to an optional mixin

kylegibson:
The base PyDocX exporter accepts an option `convert_root_level_upper_roman` which, if `True`, will convert root level lists that are formatted to "upperRoman" to a heading of level 2. 

* Decouple this logic from the pre-processor / exporter, and implement it in an optional mixin
* Remote the `convert_root_level_upper_roman` option

Depends on #135

### DIFF
--- a/pydocx/export/mixins/convert_root_upper_roman_list_to_heading.py
+++ b/pydocx/export/mixins/convert_root_upper_roman_list_to_heading.py
@@ -1,0 +1,29 @@
+# coding: utf-8
+from __future__ import (
+    absolute_import,
+    print_function,
+    unicode_literals,
+)
+
+
+class ConvertRootUpperRomanListToHeadingMixin(object):
+    def parse_p(self, el, text, stack):
+        if text == '':
+            return ''
+
+        # TODO This is being done in the base exporter already
+        justified_text = self.justification(el, text)
+
+        heading_style_name = self.pre_processor.heading_level(el)
+
+        if heading_style_name:
+            return self.heading(
+                text=justified_text,
+                heading_style_name=heading_style_name,
+            )
+
+        return super(ConvertRootUpperRomanListToHeadingMixin, self).parse_p(
+            el,
+            text,
+            stack,
+        )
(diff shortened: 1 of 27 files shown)


pydocx/export/mixins/convert_root_upper_roman_list_to_heading.py
```diff
@@ -0,0 +1,49 @@
...
+
+        return level.num_format == NumberingDefinitionsPart.NUM_FORMAT_UPPER_ROMAN  # noqa
+
+    def parse_p(self, el, text, stack):
+        if text == '':
+            return ''
+
+        if self._is_element_a_root_level_upper_roman_list_item(el):
+            # TODO justification is being done in the base exporter
+            justified_text = self.justification(el, text)
```
winhamwr: Since we're not overwriting this, should we make this a `super` call?

pydocx/openxml/packaging/numbering_definitions_part.py
```diff
@@ -22,13 +22,22 @@ class NumberingDefinitionsPart(OpenXmlPart):
...
     def numbering(self):
-        if self._numbering:
-            return self._numbering
-        self._numbering = Numbering.load(self.root_element)
+        if not self._numbering:
+            self.numbering = self.load_numbering()
         return self._numbering
+
+    @numbering.setter
+    def numbering(self, numbering):
```
winhamwr: I'm -0 on this setter. Seems pretty easy to just use `self._numbering` directly in the numbering property and to make `load_numbering` just return the numbering and for the property to be responsible for assignment.

pydocx/export/mixins/convert_root_upper_roman_list_to_heading.py
```diff
@@ -0,0 +1,49 @@
...
+    unicode_literals,
+)
+
+from pydocx.openxml.packaging import NumberingDefinitionsPart
+
+
+class ConvertRootUpperRomanListToHeadingMixin(object):
+    HEADING_LEVEL = 'heading 2'
+
+    def _is_element_a_root_level_upper_roman_list_item(self, el):
```
winhamwr: This method is surprisingly easy to read. The refactoring you've been working on is paying off in readability.

winhamwr:
:+1:

pydocx/export/mixins/convert_root_upper_roman_list_to_heading.py
kylegibson: Calling supper would do more than we want

PR #139, opened 2015-05-13
Update iterative parser to pass a context to each parsing handler

kylegibson:


michaelgzoller:
Looks good to me, @kylegibson

PR #140, opened 2015-05-17
fixed #103: incorrect row span

### LINKED ISSUE #103: Table rowspan not properly parsed
[This](https://s3.amazonaws.com/uploads.hipchat.com/6795/13776/iuEBI20v1KqYYXT/test%20table%20for%20Savannah.docx) docx file results in invalid HTML. The row with "19 to 64 yrs" in the first td comes in with `rowspan="5"` when it should come in with `rowspan="2"`.

swiperthefox:


kylegibson:
Awesome! Thank you for the pull-request. Could you include a test case?

swiperthefox:
I thought about writing a test before submitting it, but not quite sure what's the proper way to do it: to use a docx file or an xml string. 

 Here is a test case using docx file, added in test_docx.py.

kylegibson:
Perfect, thank you!

PR #141, opened 2015-05-17
Update parser to include the next_element in the context passed to parser handlers

kylegibson:


michaelgzoller:
Also looks good, @kylegibson

PR #144, opened 2015-05-21
"cString" import doesn't exist: it's a typo

kylegibson:
In several locations, we try to import from `cString` (instead of `cStringIO`). This exception is ignored because it's inside a try...except for the purpose of supporting python2/python3.


michaelgzoller:
:+1:

PR #146, opened 2015-05-21
Fix FutureWarning with testing element existence

kylegibson:
```
pydocx/pydocx/models.py:248: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if element:
pydocx/pydocx/export/base.py:318: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
  if not main_document_part.footnotes_part.root_element:
```


michaelgzoller:
:+1:

PR #145, opened 2015-05-21
Add support for modeling XML content

kylegibson:
For example,

```
<foo>
   <data>bar</data>
</foo>
```

The XML model schema should have a way of accessing "bar" in the `data` node. Currently it doesn't. We'll need this for `t` and `delText` among others.


michaelgzoller:
:+1:

PR #149, opened 2015-06-01
Remove the Convert Upper Roman Numeral mixin

kylegibson:
We're not actually using it, so remove it. See #136


michaelgzoller:
Looks good to me

PR #151, opened 2015-06-02
Implement new exporter using object classes and generators

kylegibson:
* Create OOP classes for each openxml item (Document, Body, Paragraph, Run, Text, etc)
* Rewrite the exporter to use the new classes
* Eliminate the pre-processor
* Convert the bottom up text-building export algorithm to a top-down generator algorithm that uses the HtmlTag abstraction


kylegibson:
TODO

- [ ] New version isn't handling tabs. That also means we don't have a test case on tabs.

jlward:
Can I get a small overview of changes? This is going to be tough to review without some sort of help.

kylegibson:
@jlward I added some bullet points to the description. Start at `export()` in the html exporter.

pydocx/export/html.py
```diff
@@ -7,24 +7,112 @@
...
+def get_first_from_sequence(sequence, default=None):
+    first_result = default
+    try:
+        first_result = next(sequence)
+    except StopIteration:
+        pass
+    return first_result
+
+
+class HtmlTag(object):
```
jlward: I really like this extraction. It forces us to make sure all tags are created the same way (pesky dictionary based attrs) and deals with self closing tags by default.

+1

pydocx/export/html.py
```diff
@@ -142,268 +152,711 @@ def style(self):
...
+            else result
+            for result in super(PyDocXHTMLExporter, self).export()
+        )
+
+    def export_document(self, document):
+        tag = HtmlTag('html')
+        results = super(PyDocXHTMLExporter, self).export_document(document)
+        return tag.apply(chain(self.head(), results))
+
+    def calculate_numbering_spans(self, paragraphs):
```
jlward: TODO, this is quite the method. I am going to skip reviewing this for now and come back after I have a better understanding of how all the pieces fit together.

kylegibson: Yeah, this is easily the most complicated part

pydocx/export/html.py
```diff
@@ -142,268 +152,711 @@ def style(self):
...
-
-    def deletion(self, text, author, date):
-        return self.make_element(
-            tag='span',
-            contents=text,
-            attrs={
-                'class': 'pydocx-delete',
-            },
+            attrs = {}
+            if self._is_ordered_list(level):
```
jlward: How about:

```python
if self._is_ordered_list(level):
    attrs['class'] = pydocx_class
    tag_name = 'ol'
else:
    tag_name = 'ul'
return HtmlTag(tag_name, **attrs)
```

I don't see any reason to return out of each nesting. Thoughts?

kylegibson: Definitely better

pydocx/export/html.py
```diff
@@ -142,268 +152,711 @@ def style(self):
...
+    def export_hyperlink(self, hyperlink):
+        results = super(PyDocXHTMLExporter, self).export_hyperlink(hyperlink)
+        tag = self.get_hyperlink_tag(hyperlink)
+        if tag:
+            results = tag.apply(results, allow_empty=False)
+
+        # Prevent underline style from applying by temporarily monkey-patching
+        # the export unliner function. There's got to be a better way.
+        old = self.export_run_property_underline
+        self.export_run_property_underline = lambda run, results: results
```
jlward: sad panda :(

pydocx/export/html.py
```diff
@@ -142,268 +152,711 @@ def style(self):
...
+            return HtmlTag('a', href=href)
+
+    def export_hyperlink(self, hyperlink):
+        results = super(PyDocXHTMLExporter, self).export_hyperlink(hyperlink)
+        tag = self.get_hyperlink_tag(hyperlink)
+        if tag:
+            results = tag.apply(results, allow_empty=False)
+
+        # Prevent underline style from applying by temporarily monkey-patching
+        # the export unliner function. There's got to be a better way.
```
jlward: I think unliner -> underliner? Maybe?

pydocx/export/html.py
kylegibson: Yeah, definitely interested in a better option here...

pydocx/export/html.py
kylegibson: Yep

pydocx/export/html.py
kylegibson: One alternative is the underline can detect if there's a hyperlink parent, and bail if there is.

pydocx/export/html.py
jlward: Take 2 :)

jlward: Can you add a comment or a docstring at the top of this guy that informs the reader what each of the key names mean (active, open-level, etc)? I think I have a pretty good idea, but I want to be sure. And if I could use some help here, someone who doesn't know you or pydocx will most definitely need it :)

tests/export/html/test_numbering.py
```diff
@@ -352,8 +352,250 @@ def test_list_with_empty_numbering_xml(self):
...
+        document.add(MainDocumentPart, document_xml)
+
+        expected_html = '''
+            <ol class="pydocx-list-style-type-lowerLetter">
+                <li>
+                    AAA
+                    <br />
+                    foo
+                </li>
+                <li>AAA</li>
```
jlward: I don't like that we have `AAA

kylegibson: Good point. I'll fix this

jlward:
Done with my first pass. A few things:

1. I notice we have off loaded quite a bit of the logic to the HTML subclass of the parser. I assume this is done to prevent HTML related stuff from polluting the main parser. Is this correct?
1. I don't see any changes to the markdown parser. Is it now broken? If it is, that's find, I would however like to see an issue created with plans on how to fix this problem.
1. I do want to do another pass after comments/docstrings on that big method.

jlward:
Also, can you make a comment somewhere with all the OOXML tags that do not have classes for them yet?

kylegibson:
> Also, can you make a comment somewhere with all the OOXML tags that do not have classes for them yet?

There are likely hundreds. I don't know if that would be practical.

kylegibson:
> I notice we have off loaded quite a bit of the logic to the HTML subclass of the parser. I assume this is done to prevent HTML related stuff from polluting the main parser. Is this correct?

It wasn't clear what was HTML-specific, and what wasn't. So if it was slightly ambiguous or complicated, I left it in the HTML exporter for now. Once the markdown exporter is updated, it should help reveal certain patterns for abstraction.

kylegibson:
> I don't see any changes to the markdown parser. Is it now broken?

It's broken on master. I'll make an issue.

jlward:
I think this looks good now.

kylegibson:
- [ ] Investigate adding test case to cover the rowspan bug that was fixed
- [ ] Handle `numFmt = none`
- [ ] Previous non-numbering paragraphs not handled properly when the next **different* numbering definition encountered is also a heading

kylegibson:
Notes from a discussion I had with @winhamwr over the phone regarding headings in list items:

* Current pydocx behavior breaks numbering when a heading is detected in a list item. The heading is pulled out of the list item. This functionality is maintained in this issue. There are improvements. In some cases the previous algorithm would generate invalid HTML. In others, it would still result in a heading tag in a sub-list which would then be problematic for wymeditor (see the next bullet)
* The wymeditor doesn't support h1 html header tags in list items, so we can't just do a 1-1 translation for HTML that would be consumed by it
* We suspect that most of the time people use a heading style for a paragraph in a list item, they just want the text to be emphasized or larger.
* For now I will create a pydocx issue explaining the problem

pydocx/export/base.py
```diff
@@ -6,269 +6,58 @@
...
+            wordprocessing.FootnoteReferenceMark: self.export_footnote_reference_mark,
+            wordprocessing.SdtRun: self.export_sdt_run,
+            wordprocessing.SdtContentRun: self.export_sdt_content_run,
+            wordprocessing.SdtBlock: self.export_sdt_block,
+            wordprocessing.SdtContentBlock: self.export_sdt_content_block,
+            wordprocessing.TabChar: self.export_tab_char,
+            vml.Shape: self.export_vml_shape,
+            vml.ImageData: self.export_vml_image_data,
+            NumberingSpan: self.export_numbering_span,
+            NumberingItem: self.export_numbering_item,
```
kylegibson: @jlward here

pydocx/export/base.py
```diff
@@ -298,706 +87,266 @@ def numbering_definitions_part(self):
...
+                yield result
+
+    def yield_numbering_spans(self, items):
+        builder = NumberingSpanBuilder(items)
+        numbering_spans = builder.get_numbering_spans()
+        for item in numbering_spans:
+            yield item
+
+    def export_body(self, body):
+        numbering_spans = self.yield_numbering_spans(body.children)
```
kylegibson: @jlward here

pydocx/export/base.py
```diff
@@ -298,706 +87,266 @@ def numbering_definitions_part(self):
...
+    def export_table_row(self, table_row):
+        return self.yield_nested(table_row.cells, self.export_node)
 
-    @property
-    def parsed(self):
-        if not self._parsed:
-            self._load()
-        return self._parsed
+    def export_table_cell(self, table_cell):
+        numbering_spans = self.yield_numbering_spans(table_cell.children)
```
kylegibson: @jlward here

pydocx/export/base.py
```diff
@@ -298,706 +87,266 @@ def numbering_definitions_part(self):
...
+
+    def export_tab_char(self, tab_char):
+        return
+        yield
 
-    @abstractmethod
-    def linebreak(self):
-        return ''
+    def export_numbering_span(self, numbering_span):
+        return self.yield_nested(numbering_span.children, self.export_node)
```
kylegibson: @jlward here

pydocx/export/base.py
```diff
@@ -298,706 +87,266 @@ def numbering_definitions_part(self):
...
-        self,
-        text,
-        alignment=None,
-        left=None,
-        right=None,
-        firstLine=None,
-    ):
-        return text
+    def export_numbering_item(self, numbering_item):
+        return self.yield_nested(numbering_item.children, self.export_node)
```
kylegibson: @jlward here

pydocx/export/html.py
```diff
@@ -142,268 +182,470 @@ def style(self):
...
-            },
-        )
-
-    def list_element(self, text):
-        return self.make_element(
-            tag='li',
-            contents=text,
-        )
+        for result in results:
+            yield result
```
kylegibson: @jlward HTML export_paragraph is SUPER simple now (remember the numbering_level_begin, and the line_break stuff?)

pydocx/export/html.py
```diff
@@ -142,268 +182,470 @@ def style(self):
...
-        attrs = {}
-        if col:
-            attrs['colspan'] = col
-        if row:
-            attrs['rowspan'] = row
-        return self.make_element(
-            tag='td',
-            contents=text,
-            attrs=attrs,
+    def export_numbering_span(self, numbering_span):
```
kylegibson: @jlward here

pydocx/export/html.py
```diff
@@ -142,268 +182,470 @@ def style(self):
...
-    def break_tag(self):
-        return '<br />'
+        tag_name = 'ul'
+        if not numbering_span.numbering_level.is_bullet_format():
+            attrs['class'] = pydocx_class
+            tag_name = 'ol'
+        tag = HtmlTag(tag_name, **attrs)
+        return tag.apply(results)
+
+    def yield_with_line_breaks_between_paragraphs(self, items):
```
kylegibson: @jlward here

pydocx/export/html.py
```diff
@@ -142,268 +182,470 @@ def style(self):
...
+                if isinstance(previous, wordprocessing.Paragraph) and previous.children:
+                    yield wordprocessing.Break()
+            yield item
+            previous = item
+
+    def export_numbering_item(self, numbering_item):
+        children = self.yield_with_line_breaks_between_paragraphs(numbering_item.children)
+        results = self.yield_nested(children, self.export_node)
+        tag = HtmlTag('li')
+        return tag.apply(results)
```
kylegibson: @jlward here

pydocx/export/numbering_span.py
```diff
@@ -0,0 +1,284 @@
+# coding: utf-8
```
kylegibson: @jlward this entire file

jlward:
@kylegibson still looks good :)

kylegibson:
- [ ] This still needs massive documentation updates before it can be merged in

CHANGELOG.rst
```diff
@@ -1,12 +1,14 @@
...
-- parser handlers have been updated
-  to accept a single ``context``
-  which is a namedtuple
-  consisting of
-  the previous arguments
-- Added ``next_element`` to parser context
+- Note: This release consists of significant changes to the internal API and is not
+  backwards compatible with prior versions
 - Removed ``ConvertRootUpperRomanListToHeadingMixin``
+- Fixed issue where the same image referenced multiple times would not
```
michaelgzoller: Literally the same image, or an image with the same file name?

kylegibson: Literally the same image stream, technically.

kylegibson:
@michaelgzoller Can you leave a comment here acknowledging that you've reviewed the documentation and changelog updates?

michaelgzoller:
Documentation and changelog have been reviewed and look good.

PR #168, opened 2015-07-06
ensure that the file pointer is at the beginning

IuryAlves:
When you have an image which is referenced more than one time. In the second time   `image_part.stream.read()` in the method `parse_image` returns a empty string.
I fixed it back to the beginning of the file, before reading the file. more specifically placing the following code:  `image_part.stream.seek(0)`

Here is a docx that reproduces this bug: https://drive.google.com/file/d/0B7-HjtaVhQPSamxndVo0Q00tT2M/view?usp=sharing


winhamwr:
Hello Lury,

Thanks so much for the pull request!

To speed up the process of getting this merged in, is there any way you could add a unit test for this?

You can add the document that you linked to [fixtures](https://github.com/CenterForOpenScience/pydocx/tree/master/tests/fixtures) with a correlating `.html` file of the output you expect. Then, you can add the name of the file/html to [cases](https://github.com/CenterForOpenScience/pydocx/blob/master/tests/export/test_docx.py#L25) to have the .docx file loaded and its output compared to our corresponding HTML file. 

Thanks!
-Wes

IuryAlves:
@winhamwr Done!

It takes me some time to realize that the test html is just the contents, not the whole html = D

winhamwr:
Hi Lury,

Sorry for the slow response. Your testcase looks good and hopefully @kylegibson will be along to give it a close review and merge, soon.

Thanks again!
-Wes

IuryAlves:
@winhamwr  Thanks =D

jlward:
Looks good. Thanks for the PR

Issue #159, opened 2015-06-20
Resize images to an appropriate size for the document


jhubert:
I'm dealing with a number of word documents that have a thumbnail of an image in them, but where the thumbnail is actually just a full-size image visually scaled down.

When PyDocX processes the images, it keeps the image file the same size and uses visual rules to show the image in the intended size.

Do you have any suggestions for how I would set it up to resize the actual image file so that the file size isn't so large?

winhamwr:
We actually implemented something like this in a custom attachment handler.
If I don't post it Monday, please bump this.

-Wes

jhubert:
Awesome. Thanks!

winhamwr:
Hi Jeremy,

Here is a gist for a [resizing pydocx image handler](https://gist.github.com/winhamwr/9dbf16a5a73759e35c43). It's copy/pasted from various parts of our code base, but should include most of the necessary things. It uses `PIL` to do resizing, but is smart about when to resize. It also converts some nasty bitmap formats (bmp, tiff) to compressed formats and handles base64-encoded images.

We would really appreciate it if you could turn this into a mixin in the vein of the [FakedSuperscriptAndSubscriptExportMixin](https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/export/mixins/faked_superscript_and_subscript.py#L11) that we could add to the project and then maintain. I'm sure you and I aren't the only ones that would like to solve this problem and it makes sense for it to live out in the open so we can be sure and maintain a working version.

Thanks
-Wes

jhubert:
That's great. Thanks for sharing. I'll take a crack at turning it into a Mixin. I'm not much of a python developer, but I should be able to figure it out. Nothing a code review or two can't fix. :)

jhubert:
@winhamwr Well, this was over my head as far as Python goes. I just didn't end up having the time to sit down and learn everything to implement this properly with tests in Python. I ended up hiring someone to help, and I've submitted the pull-request for review. Thanks again for sharing the original code. Feedback appreciated.

winhamwr:
Hello Jeremy,

> I ended up hiring someone to help, and I've submitted the pull-request for review. Thanks again for sharing the original code. Feedback appreciated.

Thanks so much for sharing the code. Are you using that pull-request in production somewhere?

[...]

PR #134, opened 2015-07-07
Implement faked-list detection

kylegibson:
https://github.com/kylegibson/pydocx/compare/issue_151...kylegibson:issue_134

Scenarios:

## Full-sequence nested partial-faked list using indentation, local style

1, 2, 2.1, 2.2 are "real" lists. 3, 3.1, 3.2 are faked using indentation

```
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:numPr>
          <w:ilvl w:val="0"/>
          <w:numId w:val="1"/>
        </w:numPr>
        <w:rPr/>
      </w:pPr>
      <w:r>
        <w:rPr/>
        <w:t>a</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:numPr>
          <w:ilvl w:val="0"/>
          <w:numId w:val="1"/>
        </w:numPr>
        <w:rPr/>
      </w:pPr>
      <w:r>
        <w:rPr/>
        <w:t>b</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:numPr>
          <w:ilvl w:val="1"/>
          <w:numId w:val="1"/>
        </w:numPr>
        <w:rPr/>
      </w:pPr>
      <w:r>
        <w:rPr/>
        <w:t>c</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:numPr>
          <w:ilvl w:val="1"/>
          <w:numId w:val="1"/>
        </w:numPr>
        <w:rPr/>
      </w:pPr>
      <w:r>
        <w:rPr/>
        <w:t>d</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:tabs>
          <w:tab w:val="left" w:pos="542" w:leader="none"/>
        </w:tabs>
        <w:ind w:left="0" w:right="0" w:firstLine="360"/>
        <w:rPr/>
      </w:pPr>
      <w:r>
        <w:rPr/>
        <w:t>3.   e</w:t>
      </w:r>
    </w:p>
    <w:p>
      <w:pPr>
        <w:pStyle w:val="Normal"/>
        <w:tabs>
          <w:tab w:val="left" w:pos="542" w:leader="none"/>
[...]

PR #169, opened 2015-07-08
Add a mixin for resizing images on export

jhubert:
The PyDocXHTMLExporterImageResizeMixin mixin resizes larger images
down to the appropriate size for them to be displayed in the resulting
HTML. It doesn't resize images that are the same size or smaller than
the scaled placement in the docx file.

Based on the conversation in CenterForOpenScience/pydocx#159


pydocx/export/image_resize.py
```diff
@@ -0,0 +1,165 @@
...
+    print_function,
+    unicode_literals,
+)
+
+import base64
+from StringIO import StringIO
+from urlparse import urlparse
+import posixpath
+
+import requests
```
winhamwr: This needs to be added to [install_requires](https://github.com/CenterForOpenScience/pydocx/blob/master/setup.py#L54).

pydocx/export/image_resize.py
```diff
@@ -0,0 +1,165 @@
...
+
+from requests.exceptions import InvalidSchema
+
+from pydocx.util import uri
+from PIL import Image
```
winhamwr: This needs to be added to [install_requires](https://github.com/CenterForOpenScience/pydocx/blob/master/setup.py#L54), or otherwise communicated. Ideally, we wouldn't require a working PIL installation if someone didn't want to use this mixin. Thoughts @kylegibson @jlward ?

jhubert: Yeah, I agree. I wasn't sure how you guys wanted this kept separate from the main code. Thought I'd send up the pull-request as-is for us to discuss. Happy to move things around to make it all optional.

However, as I was working on this, I struggled to think of a time when you **wouldn't** want the large images resized. Since images are being converted to DataURIs, it seems rather silly to have large images that are scaled down be kept at full size.

Can you help me think of some scenarios where you wouldn't want the behavior?

winhamwr: > However, as I was working on this, I struggled to think of a time when you wouldn't want the large images resized. 

I can't really think of one, either. I'm +1 on making this the default behavior, even.

setup.py
```diff
@@ -51,7 +51,7 @@ def main():
         },
         scripts=[],
         zip_safe=False,
-        install_requires=[],
+        install_requires=['requests', 'PIL'],
```
jlward: We have been very careful not to force any required packages for this project (thusly making it as slim as possible). I would much rather say that these are only required if you want to use this new mixin. Thoughts?

jlward:
Can you take a look at the travis tests to figure out why all of them are failing? I don't foresee us merging in an issue that has failing tests.

setup.py
jhubert: I'm more than happy to have them as optional. Unfortunately I'm not good enough at Python to understand how you would do that in this project. In the other mixin, I've just made it an entirely separate python library and that keeps everything out of this project, but I'm very open to suggestions if you have a good idea of how to do that and still keep this resizer in the lib.

@winhamwr mentioned that resizing might actually be worth including as default behavior, so I'm not sure how that would affect this decision.

setup.py
```diff
@@ -51,7 +51,7 @@ def main():
         },
         scripts=[],
         zip_safe=False,
-        install_requires=[],
+        install_requires=['requests', 'Pillow'],
```
jlward: If you checkout how we are handling the optional requirement defusedxml, that should give you a great place to start. The key is, in your mixin, you'll want to do something like:

```python
try
    import requests
    from PIL import Image
except ImportError:
    use_resize_mixin = False
else:
    user_resize_mixin = True
```

Then, at the entry point for the mixin check to see what the value of that boolean flag is and early exit if those two libraries are missing. Does that make sense?

I would be more than happy to help you if you would like. I'm pretty excited about this mixin.

jhubert: Awesome. Thanks for the tip in the right direction and I'm glad you're excited! :grinning: 

I was thinking about how to keep everything isolated so that the mixin's source doesn't pollute the main source code. That's one of the reasons https://github.com/jhubert/pydocx-s3-images was done as a standalone package.

I am not going to have time to work on this until Wednesday, but I've added you to my fork if you want to take a look / make any commits. I'd love the help. :construction_worker: 

I'm using this in production right now and it's made a huge improvement already. The file sizes I'm getting back are much more reasonable.

jhubert:
Odd, the failing test is a `test performance` test. The suite takes 7.4s on an old architecture but the test wants 7 seconds.

kylegibson:
Yeah that test case fails frequently. We need a better way to measure "performance" than time, since slower systems will naturally be slower so we can't rely on a fixed amount of time. I've been tempted to rip the test out, but it is useful to a degree, and I don't know of a better way.

kylegibson:
Also, a few hours ago I merged a rather significant change into master that will likely conflict with the changes you've made in this PR. I apologise for that. Once I have #134 documented and also merged in, I can help you fix any conflicts here.

jhubert:
@kylegibson @jlward This has been extracted into a standalone plugin instead of trying to fit it in to the PyDocX project.

https://github.com/jhubert/pydocx-resize-images

Closing this pull request, but happy to look at other options.

PR #162, opened 2015-08-19
Don't break numbering when handling headings inside list items

kylegibson:
Currently, pydocx will break numbering for headings inside list items. For example:

1. **Foo**
2. Bar
3. **Baz**

If Foo and Baz are headings, the resulting HTML would look something like:

# Foo
1. Bar
# Baz

The rationale for doing this was to avoid list items that contained HTML header tags, i.e.

```html
<ol>
<li><h1>Foo</h1></li>
</ol>
```

## Moving forward

* Stop breaking numbering. We should mirror, as best as possible, what Word displays.
* Ignore headings (but preserve the styles derived from headings) in list items.
* Instead of applying header tags, possibly force-apply bolding (even if the style doesn't include it)


tests/export/html/test_heading.py
```diff
@@ -340,9 +348,16 @@ def test_heading_in_list_with_bare_paragraph(self):
...
 
         expected_html = '''
-            <h1>foo</h1>
-            <p>bare paragraph</p>
-            <h1>bar</h1>
+            <ol class="pydocx-list-style-type-decimal">
+                <li>
+                    <strong>foo</strong>
+                    <br />
+                    bare paragraph
```
jlward: I don't understand why this is happening. Shouldn't this be two lists separated by a paragraph? That is closer to what the XML is showing happening.

tests/export/html/test_heading.py
```diff
@@ -506,10 +523,9 @@ def test_heading_as_list_following_bare_paragraph_plus_list(self):
 
         expected_html = '''
             <ol class="pydocx-list-style-type-decimal">
-                <li>foo</li>
+                <li>foo<br />bare paragraph</li>
```
jlward: Same thing here. I assume we are rolling them in because they have the same `numId`

jlward:
Assuming my assessment of the two previous comments is correct, then this guy is ready to go. If I have actually caught a bug, then do your thing and ping me for another review.

tests/export/html/test_heading.py
kylegibson: Since the numId is the same, it's the same list.

tests/export/html/test_heading.py
kylegibson: Yep, that's how numIds work.

jlward: In hindsight, I do recall building that functionality at some point.

PR #163, opened 2015-08-25
Handle numFmt = none

kylegibson:
It's possible for `numFmt` to be `none`. In this case, don't display any list styling.


michaelgzoller:
:+1:  @kylegibson

PR #178, opened 2015-09-02
Headings should preserve certain styles

kylegibson:
Currently, headings strip out all styling information. This prevents weird cases like `<h2><strong>Foo</strong></h2>` but it also means that headings don't have any other styles, like italics. Update headings to preserve italic. Also, headings that use vanish or hidden styles should have the vanished or hidden spans applied.


jlward:
Looks good

PR #161, opened 2015-09-03
Styled whitespace is ignored

kylegibson:
Actual result: `<p>FooBar</p>`

```python
    def test_single_styled_whitespace_in_text_run_is_preserved(self):
        document_xml = '''
            <p>
              <r>
                <t>Foo</t>
              </r>
              <r>
                <rPr>
                    <b />
                </rPr>
                <t> </t>
              </r>
              <r>
                <t>Bar</t>
              </r>
            </p>
        '''
        document = WordprocessingDocumentFactory()
        document.add(MainDocumentPart, document_xml)

        expected_html = '<p>Foo Bar</p>'
```

### DIFF
--- a/tests/export/html/test_paragraph.py
+++ b/tests/export/html/test_paragraph.py
@@ -11,6 +11,50 @@ from pydocx.test.utils import WordprocessingDocumentFactory
 from pydocx.openxml.packaging import MainDocumentPart
 
 
+class NoEmptyParagraphsTestCase(DocumentGeneratorTestCase):
+    def test_no_runs_no_text(self):
+        document_xml = '<p></p>'
+        document = WordprocessingDocumentFactory()
+        document.add(MainDocumentPart, document_xml)
+
+        expected_html = ''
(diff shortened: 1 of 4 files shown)


tests/export/html/test_paragraph.py
```diff
@@ -11,7 +11,133 @@
...
+                <rPr>
+                  <b />
+                  <i />
+                  <u val="single"/>
+                  <caps />
+                  <smallCaps />
+                  <dstrike />
+                  <strike />
+                  <webHidden />
+                  <vanish />
```
michaelgzoller: `vanish` doesn't make it vanish?

michaelgzoller:
:+1:

tests/export/html/test_paragraph.py
kylegibson: The default is to surround the vanished content with a span that has the pydocx-vanish class. You can then customize the behavior in CSS.

PR #181, opened 2015-09-21
Don't include head if it is None

kylegibson:
The current implementation requires that `head` in the html exporter return an iterable. If it returns `None`, you have a bad time.


michaelgzoller:
:+1:

PR #185, opened 2015-09-23
Multiple complex fields sharing the same parent causes content loss

kylegibson:
If a paragraph contains two or more complex fields, content disappears.


tests/export/html/test_field_code.py
```diff
@@ -12,6 +12,86 @@
...
+                    <fldChar fldCharType="begin"/>
+                </r>
+                <r>
+                    <instrText> FOOBAR baz</instrText>
+                </r>
+                <r>
+                    <fldChar fldCharType="separate"/>
+                </r>
+                <r>
+                    <t>BBB</t>
```
michaelgzoller: So the behavior before was BBB would disappear?

michaelgzoller:
What happens if multiple instruction fields are nested? Is that even possible?

michaelgzoller:
:+1:

tests/export/html/test_field_code.py
kylegibson: No, actually, CCC would disappear.

kylegibson:
> What happens if multiple instruction fields are nested? Is that even possible?

Everything breaks. It's only possible manually. The spec doesn't allow that.

PR #188, opened 2015-09-23
Heading inside complex field has strong applied

kylegibson:
For example, the current output is:

```html
<h2><strong>Foo</strong></h2>
```

The desired output is

```html
<h2>Foo</h2>
```

This still works for non-complex field runs but is currently broken for complex field runs.

### DIFF
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -180,6 +180,8 @@ class PyDocXExporter(object):
                     previous_parent_new_children.append(child)
             first_run.parent.children = previous_parent_new_children
 
+            field.parent = first_run.parent
+
             for run in field.children:
                 run.parent = field
 
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -110,15 +110,23 @@ class PyDocXExporter(object):
             # document (e.g. fields)
             # In the first pass, discard any generated results
             self.first_pass = True
-            for result in self.export_node(document):
-                pass
+            self._first_pass_export()
 
-            self._convert_complex_fields_into_simple_fields()
+            self._post_first_pass_processing()
 
             # actually render the results
             self.first_pass = False
             for result in self.export_node(document):
                 yield result
+
+    def _first_pass_export(self):
+        document = self.main_document_part.document
+        if document:
+            for result in self.export_node(document):
+                pass
+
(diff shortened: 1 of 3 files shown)


tests/export/html/test_field_code.py
```diff
@@ -7,11 +7,58 @@
 )
 
 
-from pydocx.openxml.packaging import MainDocumentPart
+from pydocx.openxml.packaging import MainDocumentPart, StyleDefinitionsPart
 from pydocx.test import DocumentGeneratorTestCase
 from pydocx.test.utils import WordprocessingDocumentFactory
 
 
+class HeadingTestCase(DocumentGeneratorTestCase):
```
winhamwr: You have a testcase named `HyperlinkFieldCodeTestCase` in this module, so wouldn't calling this `HeadingFieldCodeTestCase` be more consistent?

kylegibson: Maybe I should rename the other class to drop the FieldCode, since the module is test_field_code?

pydocx/export/base.py
```diff
@@ -180,6 +180,8 @@ def _convert_complex_fields_into_simple_fields(self):
                     previous_parent_new_children.append(child)
             first_run.parent.children = previous_parent_new_children
 
+            field.parent = first_run.parent
```
winhamwr: I don't understand this at all. Could it benefit from a code comment, perhaps?

winhamwr:
Looks good, other than the test name and code comment suggestion.

winhamwr:
Also, I'm pretty sure this will save cleaners a lot of time. I hear often about how it's a pain to de-bold things if it comes in as a heading and it shouldn't be.

kylegibson:
> I'm pretty sure this will save cleaners a lot of time

This isn't a bug that is currently present in PolicyStat. This was introduced in v0.9.0 which hasn't gone live for us yet.

pydocx/export/base.py
kylegibson: The problem was field.parent was never being set, so it was None. Yet the new field was being inserted into the structure. As a consequence, it mean that the contained runs had no way to detect whether they were contained within a paragraph that was a heading, since the structure only went up as far as the field.

PR #190, opened 2015-09-24
Fake nested lists not converted inside tables

kylegibson:


winhamwr:
:+1: 

This bug was only introduced when the complex hyperlink fix was introduced with 0.9.0, right?

kylegibson:
That's correct

PR #209, opened 2016-04-23
office Open xml: shared_math

IuryAlves:
Hi,  I spent some time writing the tags that is used with equations. This is not complete yet, but maybe is a good start to create a docx to latex exporter.

I also noticed that there is a issue (#49) to enable equations, but it is stopped for a long time. I would like to continue contributing to make this possible.

### DIFF
--- a/pydocx/openxml/shared_math/__init__.py
+++ b/pydocx/openxml/shared_math/__init__.py
@@ -1,0 +1,76 @@
+from pydocx.openxml.shared_math.accent import Accent
+from pydocx.openxml.shared_math.accent_properties import AccentProperties
+from pydocx.openxml.shared_math.align import Align
+from pydocx.openxml.shared_math.align_scripts import AlignScripts
+from pydocx.openxml.shared_math.argument_properties import ArgumentProperties
+from pydocx.openxml.shared_math.argument_size import ArgumentSize
+from pydocx.openxml.shared_math.bar import Bar
+from pydocx.openxml.shared_math.bar_properties import BarProperties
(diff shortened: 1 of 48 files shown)


winhamwr:
Hello iury,

Thanks for the PR! We would love to have support for equations.

Would you be willing to implement at least some of these models in an exporter (preferably HTML), or perhaps write a separate exporter that adds support for these? We'd also love to have some basic smoke tests covering a .docx file with a mix of math and normal content, along with some XML tests covering all of the added models. It doesn't need to cover all of the corner cases or achieve full coverage, but in order to maintain this going forward, we need at least "happy path" coverage.

Thanks again!
-Wes

IuryAlves:
@winhamwr sure. I will do it.

IuryAlves:
@winhamwr Sorry for the delay. I've come to realize that there are many cyclic import issues in my pull request. For example:

```
<m:acc>
  <m:e>
    <m:acc/>
  </m:e>
</m:acc>
```

This becomes the following:


```python
# accent.py
from pydocx.openxml.shared_math.element import Element
```

```python
# element.py
from pydocx.openxml.shared_math.accent import Accent
```

I am wondering what is the best solution for this problem.

winhamwr:
@IuryAlves you can use a string to avoid the circular import, kind of like with Django's model syntax. [here](https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/openxml/vml/rect.py#L16) is an example.

IuryAlves:
Thanks @winhamwr

Issue #210, opened 2016-05-11
Add support for colorized text


jhubert:
Colorized text is currently ignored.

![image](https://cloud.githubusercontent.com/assets/3384/15170644/826cdab8-16fc-11e6-8c67-8e63b1de26a2.png)

ends up like

```html
<p>AAA</p>
<p>BBB</p>
```

[styled_color.docx](https://github.com/CenterForOpenScience/pydocx/files/258607/styled_color.docx)

I think it would be good to support colorized text. Basically anything other than `#000000`

PR #211, opened 2016-05-11
Add support for colorized text

jhubert:
Connects #210 

It seems like this may have been intentionally left out, but it's come up a number of times for us so I thought I'd take a crack at it.

This is a very simple handler for the `<w:color />` tag that applies the color directly to a `span` tag that wraps the text. It's following the patterns I saw in the code, but [the guard clause on line 478 in html.py](https://github.com/CenterForOpenScience/pydocx/compare/master...jhubert:210-colorized-text?expand=1#diff-1e8de747886c45532983f883b63b0b84R478) feels a little out of place. If there is a better way to handle `run.properties` not being set, I'd love to hear it.

The word document used in the tests looks like this:

image


jlward:
Hello,

Your code looks good. Do you mind adding in a change log entry showcasing your changes? And feel free to add yourself to the Authors file if you'd like. After that I don't see any reason not to merge this guy in.

kylegibson:
When we upgrade our version, we'll need to disable the color handling.

jlward:
> When we upgrade our version, we'll need to disable the color handling.

I would agree. I already look at our code and know where the change needs to be done.

jhubert:
> When we upgrade our version, we'll need to disable the color handling.

@kylegibson Any easy way to make a handler like this configurable so that it can be optional for other people? It seems like some people may want the color to be stripped.

jhubert:
@jlward Will do! Should I version bump it to 0.9.7 as well?

jlward:
> Will do! Should I version bump it to 0.9.7 as well?

@jhubert Don't worry about bumping the version. I am going to try to get #203 and #204 live before we push out another version. So I'll deal with the version stuff.

jlward:
> Any easy way to make a handler like this configurable so that it can be optional for other people? It seems like some people may want the color to be stripped.

With out use case, we have to subclass the exporter anyway. So it's pretty easy to overwrite it to do nothing. I'm not saying it's the best solution :). But it does work for us.

jhubert:
@jlward Ok great. Changelog entry added. Thanks!

jhubert:
Heh. Travis / github won't pass the build if you use [ci skip]. Good to know. Re-pushing to waste cpu cycles 🍷

jlward:
Thanks for the PR :) I am going to try to get #203 and #204 fixed before we make a new version.

jhubert:
Awesome. Thanks for merging!

PR #203, opened 2016-05-11
Add support for textbox -> txbxContent

jlward:
```
      <w:r>
            <w:pict>
                <v:textbox>
                  <w:txbxContent>
                    <w:p>
                      <w:r>
                        <w:t>Foo bar baz</w:t>
                      </w:r>
                    </w:p>
                  </w:txbxContent>
                </v:textbox>
              </v:rect>
            </w:pict>
      </w:r>
```


pydocx/openxml/markup_compatibility/alternate_content.py
```diff
@@ -0,0 +1,14 @@
...
+    unicode_literals,
+)
+
+from pydocx.models import XmlModel, XmlCollection
+from pydocx.openxml.markup_compatibility.fallback import Fallback
+
+
+class AlternateContent(XmlModel):
+    XML_TAG = 'AlternateContent'
+    children = XmlCollection(Fallback)
```
jlward: I might make a pass and update all the `XmlCollection` stuff that I touched to use strings instead of classes.

pydocx/openxml/markup_compatibility/fallback.py
```diff
@@ -0,0 +1,14 @@
...
+    unicode_literals,
+)
+
+from pydocx.models import XmlModel, XmlCollection
+from pydocx.openxml.wordprocessing.picture import Picture
+
+
+class Fallback(XmlModel):
+    XML_TAG = 'Fallback'
+    children = XmlCollection(Picture)
```
jlward: For the Markup stuff I'll handle making sure they are "complete" in #204

tests/export/test_docx.py
```diff
@@ -24,34 +24,35 @@ def convert(path, *args, **kwargs):
 
 class ConvertDocxToHtmlTestCase(DocXFixtureTestCaseFactory):
     cases = (
-        'read_same_image_multiple_times',
         'all_configured_styles',
+        'export_from_googledocs',
```
jlward: Most of this is alphabetizing.

tests/export/test_docx.py
```diff
@@ -24,34 +24,35 @@ def convert(path, *args, **kwargs):
...
         'shift_enter',
         'simple',
         'simple_lists',
         'simple_table',
         'special_chars',
         'styled_bolding',
         'table_col_row_span',
+        'table_with_multi_rowspan',
         'tables_in_lists',
+        'textbox',
```
jlward: This one is new.

tests/fixtures/textbox.html
```diff
@@ -0,0 +1,5 @@
+<p>
```
jlward: Fun fact, if you look at the XML for this, it is actually nested paragraphs. Not really sure how to stop it from doing the outside one. Or even if that will be the correct solution in the general case.

tox.ini
```diff
@@ -12,6 +12,7 @@ commands =
 deps =
   -rrequirements/testing.txt
   defusedxml: defusedxml==0.4.1
+  py26: importlib
```
jlward: :( Might be time to drop support for python 2.6

pydocx/openxml/wordprocessing/textbox.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class TxBxContent(XmlModel):
+    XML_TAG = 'txbxContent'
+    children = XmlCollection(
+        'wordprocessing.Paragraph',
+    )
+
+
+class Textbox(XmlModel):
```
jlward: This guy should really live in `vml` instead of `wordprocessing` :( I'll handle this after the code review.

pydocx/models.py
```diff
@@ -120,10 +121,25 @@ class ParkingLot(XmlModel):
...
         super(XmlCollection, self).__init__(self, default=default)
-        self.types = set(types)
+        self._types = types
         self._name_to_type_map = None
 
     @property
+    def types(self):
+        return set(self._set_types(*self._types))
+
+    def _set_types(self, *types):
```
kylegibson: Do we want to consider caching this result? Would be expensive to have to run it more than once, if stuff is needing to be imported.

pydocx/models.py
```diff
@@ -120,10 +121,25 @@ class ParkingLot(XmlModel):
...
+        base_path = 'pydocx.openxml.{0}'
+        for _type in types:
+            try:
+                path, klass, = _type.rsplit('.', 1)
```
kylegibson: A stray `,` ?

pydocx/models.py
jlward: I had considered it. Do you recall if pydocx has a built in caching tool? Or should I do it the old fashioned way?

pydocx/models.py
jlward: :(

pydocx/openxml/wordprocessing/run.py
```diff
@@ -29,18 +17,19 @@ class Run(XmlModel):
...
+        'wordprocessing.NoBreakHyphen',
+        'wordprocessing.Text',
+        'wordprocessing.Drawing',
+        'wordprocessing.Picture',
+        'wordprocessing.DeletedText',
+        'wordprocessing.FootnoteReference',
+        'wordprocessing.FootnoteReferenceMark',
+        'wordprocessing.FieldChar',
+        'wordprocessing.FieldCode',
+        'markup_compatibility.AlternateContent',
```
kylegibson: Are you doing this just to eliminate the import clutter?

jlward: Yup

pydocx/models.py
jlward: A thought. This gets called for each instance. I might need to make `types` be a class method and cache it there.

tests/export/test_docx.py
kylegibson: IMO for ourselves, we should avoid adding test cases via this method and prefer adding explicit XML test cases. Being able to test against a test file is good for new developers, since it eliminates some of the overhead in getting an issue fixed. It's been by goal to gradually reduce this list down by converting them into explicit XML test cases.

pydocx/models.py
jlward: I don't think a class method would actually fix this.

tests/fixtures/textbox.html
kylegibson: This is probably going to be bad. I think we need to figure out a solution for this. I suspect we will also want additional test cases that consist of having text within the parent level paragraph as well. For example:

```
<p>
<r><t>Foo</t></r>
<r> ... (textbox) </r>
<r><t>Bar</t></r>
</p>
```

IMO this needs to be de-normalized into three separate paragraphs for HTML purposes.

pydocx/openxml/wordprocessing/run.py
kylegibson: Okay, one concern I have about this is that we're not validating the types until the class is actually used, which means errors could be hidden. Before, that validation would occur at initial run time.

tests/export/test_docx.py
jlward: I would disagree. Way back when, I wanted to have an actual docx file that was being tested for each case we were working on. It's very easy to build an XML snippet that can't actually exist in a docx file. These are smoke tests.

tests/fixtures/textbox.html
jlward: I'll see what I can do, but I don't know if I know how to do something like this anymore.

tests/export/test_docx.py
kylegibson: It's not scalable to use full documents for each specific case. You will end up with thousands of poorly named, poorly organized files.

jlward: I'm not trying to say each docx file has one and only one thing it's testing. I want to make sure that newly considered nodes exist in a docx file that we are parsing in a test. This ensures that, at least when the new node is first added, that we have a real life example that we know now works. I feel pretty strongly about having each of the nodes being processed in our tests from docx files. Mostly because back when I would fake it (in XML), I created more than a few instances of XML that could not exist in an actual docx. Meaning I was writing fixes for things that could not happen. Meaning I quite likely didn't solve the problem.

pydocx/models.py
kylegibson: https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/util/memoize.py

pydocx/openxml/wordprocessing/run.py
jlward: As long as we have a test on each of these nodes, it doesn't matter if it happens at run time vs compile time.

kylegibson: Could you verify that is the case? We'll also need to ensure that any future additions have the requisite test cases as well.

jlward:
```
  File "/home/jason/third_party_apps/pydocx/tests/test_main.py", line 70, in test_file_handles_to_docx_are_released
    result = main(['--html', input_docx.name, output.name])
  File "/home/jason/third_party_apps/pydocx/pydocx/__main__.py", line 44, in main
    return convert(output_type, docx_path, output_path)
  File "/home/jason/third_party_apps/pydocx/pydocx/__main__.py", line 15, in convert
    output = PyDocX.to_html(docx_path)
  File "/home/jason/third_party_apps/pydocx/pydocx/pydocx.py", line 13, in to_html
    return PyDocXHTMLExporter(path_or_stream).export()
  File "/home/jason/third_party_apps/pydocx/pydocx/export/html.py", line 211, in export
  File "/home/jason/third_party_apps/pydocx/pydocx/export/html.py", line 209, in <genexpr>
  File "/home/jason/third_party_apps/pydocx/pydocx/export/base.py", line 110, in export
    document = self.main_document_part.document
  File "/home/jason/third_party_apps/pydocx/pydocx/openxml/packaging/main_document_part.py", line 49, in document
    self._document = self.load_document()
  File "/home/jason/third_party_apps/pydocx/pydocx/openxml/packaging/main_document_part.py", line 53, in load_document
    self._document = Document.load(self.root_element, container=self)
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 352, in load
    kwargs[field_name] = handler(child)
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 310, in child_handler
    return field.type.load(value, **load_kwargs)
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 370, in load
    item = handler(child, **load_kwargs)
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 370, in load
    item = handler(child, **load_kwargs)
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 339, in load
    for tag_name in field.name_to_type_map.keys():
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 146, in name_to_type_map
    for type_spec in self.types:
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 129, in types
    return set(self._set_types(*self._types))
  File "/home/jason/third_party_apps/pydocx/pydocx/models.py", line 139, in _set_types
    module = importlib.import_module(base_path.format(path))
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named rordprocessing

======================================================================
FAIL: test_cli_convert_to_html_status_code (tests.test_main.MainTestCase)
----------------------------------------------------------------------
  File "/home/jason/third_party_apps/pydocx/tests/test_main.py", line 99, in test_cli_convert_to_html_status_code
    self.assertEqual(result, 0)
AssertionError: 1 != 0

----------------------------------------------------------------------
Ran 502 tests in 1.186s

FAILED (SKIP=4, errors=326, failures=1)
```

tests/export/html/test_markup_compatibility.py
```diff
@@ -0,0 +1,91 @@
...
+        expected_html = '''
+            <p>
+                <p>
+                    AAA
+                </p>
+            </p>
+        '''
+        self.assert_document_generates_html(document, expected_html)
+
+    def test_textbox_with_content_outside_of_textbox(self):
```
jlward: Continuation of https://github.com/CenterForOpenScience/pydocx/pull/203/files#r63035777 I have the tests written. However I don't have any idea what I need to do to break this up into multiple paragraphs. From what I can tell, the outside opening p tag has already happened. Which means I'll need to close it, do my thing, then re-open it. If this is the case, I have no idea how it can be done inline. I could probably build a post-processor, but it looks like we don't have those anymore.

tests/export/html/test_markup_compatibility.py
```diff
@@ -0,0 +1,91 @@
...
+        document_xml = '''
+            <p>
+                <r><t>AAA</t></r>
+                <r>
```
kylegibson: This run could conceivably have a `<t>` node both before the `alternatecontent` and after

tests/export/html/test_markup_compatibility.py
kylegibson: We're doing something like this elsewhere in the code. I'm not certain where.

jlward: I would almost rather push this live as is, and deal with it later. But that's mostly because I don't have a good solution in mind.

kylegibson: For ComplexFields, we had to implement a two-pass approach: https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/export/base.py#L115

kylegibson: What happens in the editor if you get nested paragraph tags like this?

jlward: CKEditor actually fixes it for us.

```<p>AAA<p>BBB</p></p> --> <p>AAA</p><BBB</p>```

File: pydocx/openxml/markup_compatibility/fallback.py [hunk 13]
```diff
@@ -0,0 +1,13 @@
...
+    print_function,
+    unicode_literals,
+)
+
+from pydocx.models import XmlModel, XmlCollection
+
+
+class Fallback(XmlModel):
+    XML_TAG = 'Fallback'
+    children = XmlCollection('wordprocessing.Picture')
```
kylegibson: Do you know what other child types Fallback supports?

tests/export/html/test_markup_compatibility.py
jlward: The legacy editor actually also fixes this auto-magically.

kylegibson:
Do you have any test cases for when textbox appears without the markup_compatibility? This would happen if you had a word document created in an older version of word that didn't support the markup compatibility syntax.

pydocx/openxml/markup_compatibility/fallback.py
jlward: I will be dealing with that in #204 as I mentioned during stand up :)

I don't plan to push a new release until both this and 203 are done.

jlward:
> Do you have any test cases for when textbox appears without the markup_compatibility? This would happen if you had a word document created in an older version of word that didn't support the markup compatibility syntax.

I do not. I only have it with markup_compatibility because that was the only way I could get it to happen using libreoffice.

kylegibson:
> I only have it with markup_compatibility because that was the only way I could get it to happen using libreoffice.

What version of libreoffice?

jlward:
> What version of libreoffice?

5.1. And I don't have access to 4.x

kylegibson:
> 5.1. And I don't have access to 4.x 

Our vagrant boxes ship with 3.5. Although, ideally we would create the documents using an older version of Word. I wish we had access to Word VMs like we do with IE VMs.

jlward:
> Our vagrant boxes ship with 3.5. Although, ideally we would create the documents using an older version of Word. I wish we had access to Word VMs like we do with IE VMs.

Using 3.5 and saving as a docx actually removes the contents of the docx. So there's that.

pydocx/models.py
jlward: My research leads me to believe that no caching is needed (it won't actually change anything).

pydocx/export/base.py
```diff
@@ -15,7 +15,7 @@
     NumberingSpan,
     NumberingSpanBuilder,
 )
-from pydocx.openxml import wordprocessing, vml
+from pydocx.openxml import wordprocessing, vml, markup_compatibility
```
winhamwr: alphabetize?

pydocx/export/base.py
```diff
@@ -535,3 +539,16 @@ def export_field_char(self, field_char):
...
         pass
+
+    def export_textbox(self, textbox):
+        return self.yield_nested(textbox.children, self.export_node)
+
+    def export_textbox_content(self, textbox_content):
+        return self.yield_nested(textbox_content.children, self.export_node)
+
+    # Markup Compatibility exporters
+    def export_alternate_content(self, alternate_content):
```
winhamwr: Could we call this `export_markup_compatibility_alternative_content`? I'd rather have a longer-than-ideal name that explains the concept than a name that needs a comment.

pydocx/export/base.py
```diff
@@ -535,3 +539,16 @@ def export_field_char(self, field_char):
...
+        return self.yield_nested(alternate_content.children, self.export_node)
+
+    def export_fallback(self, fallback):
```
winhamwr: `export_markup_compatibility_fallback`?

pydocx/models.py
winhamwr: Could you clarify what you researched, Jason?

pydocx/openxml/markup_compatibility/fallback.py
winhamwr: Can you add this comment?
` #TODO in #204- actually include all of the children defined by the spec`

pydocx/models.py
jlward: I printed the types in this method and could only get that printing the first time it was used in a test. The second test did not show any of the printing.

pydocx/models.py
```diff
@@ -120,10 +121,25 @@ class ParkingLot(XmlModel):
...
+        self._types = types
         self._name_to_type_map = None
 
     @property
+    def types(self):
+        return set(self._set_types(*self._types))
+
+    def _set_types(self, *types):
+        base_path = 'pydocx.openxml.{0}'
+        for _type in types:
```
winhamwr: I really don't like moving all of the imports to be string-based. That removes some of the developer-friendly niceness of immediately knowing when you typo'd an import. Django's approach of mostly using actual imports, but supporting string-based imports where necessary, seems like the best of both worlds. We use strings when we need it, and get immediate feedback on import fails for all of the other cases.

pydocx/openxml/wordprocessing/run.py
winhamwr: I don't think testcase coverage is sufficient to mitigate that concern. We're not the only users of PyDocx. We also need to think about the developer experience for other folks that will be using and extending the library.

tests/export/html/test_markup_compatibility.py
winhamwr: Having an editor that fixes it for us is good, but other users of pydocx might not have that luxury. Our goal is to output clean HTML, right? I'm +0 on fixing this in a separate ticket, before #204 or as part of #204, if you think that would be a better way to structure the work.

tests/export/html/test_markup_compatibility.py
winhamwr: Kyle: Are you saying we need a testcase to cover that?

tox.ini
winhamwr: +1 on dropping python 2.6 support if it makes anything easier. We could do that in this PR or in a separate PR and then bump the major version for the next release.

tests/export/html/test_markup_compatibility.py
kylegibson: > Kyle: Are you saying we need a testcase to cover that?

If the standard says it is possible, then we should at least peg the behavior, and possible implement corrective actions if the existing behavior is subpar.

jlward: I updated the test to cover this. This diff is just out of date.

tests/export/html/test_markup_compatibility.py
jlward: I don't have a good way of fixing it. Honestly, I was going to punt on it, get #204 done, then not worry about it until someone finds a good way to fix it.

pydocx/openxml/wordprocessing/run.py
jlward: I don't know how to solve your problem then. If we want to allow strings (which I am +1 on), then we have to rely on tests. If we don't want strings, then we have to update things in odd places.

pydocx/export/base.py
```diff
@@ -535,3 +539,16 @@ def export_field_char(self, field_char):
...
     def export_field_code(self, field_code):
         pass
+
+    def export_textbox(self, textbox):
+        return self.yield_nested(textbox.children, self.export_node)
+
+    def export_textbox_content(self, textbox_content):
+        return self.yield_nested(textbox_content.children, self.export_node)
+
+    # Markup Compatibility exporters
```
winhamwr: Seems like we no longer need this comment.

winhamwr:
Everything looks good now except for the paragraph nesting thing. It would be good to include Kyle in that decision, since I feel like I don't have a great understanding of what it would take to implement a fix.

winhamwr:
I'll wait to review the description of the ticket to fix the nested paragraph before giving this one the ol' :+1:

jlward:
> I'll wait to review the description of the ticket to fix the nested paragraph before giving this one the ol' :+1:

#213 has been created to deal with nested paragraphs.

caffodian:
@kylegibson has reviewed #213 so this is good

Issue #195, opened 2015-11-17
Document pydocx-resize-images plugin


winhamwr:
Our documentation should include a resources section that promotes and links to [pydocx-resize-images](https://github.com/jhubert/pydocx-resize-images).

winhamwr:
Related to #159

jhubert:
@winhamwr Agree. Also, happy to roll the code from pydocx-resize-images into this project. I only released it as a plugin so that I wasn't held back by the review / release cycle.

The https://github.com/jhubert/pydocx-s3-images makes more sense as a plugin though. 👯

jhubert:
@winhamwr docs have been merged in. Want to close this issue? 😄

winhamwr:
Thanks for the reminder! Hooray docs.

PR #212, opened 2016-05-13
Add a plugins section to the documentation

jhubert:
This is a quick pass at some documentation for the third party plugins I wrote. I tried using the language that the rest of the documentation uses, but I may have injected a bit more personality than intended. Very open to rewrites; Just wanted to get something going.

Per #195  /cc @winhamwr 

The resulting documentation page looks like this:

image


kylegibson:
Looks great! Thank you!

PR #204, opened 2016-05-16
Add support for the markup compatibility namespace

jlward:
mc = http://schemas.openxmlformats.org/markup-compatibility/2006

Example:

```
      <w:r>
        <mc:AlternateContent>
          <mc:Choice Requires="wps">
            <w:drawing>
              <wp:anchor behindDoc="0" distT="0" distB="0" distL="114935" distR="114935" simplePos="0" locked="0" layoutInCell="1" allowOverlap="1" relativeHeight="2">
                <wp:simplePos x="0" y="0"/>
                <wp:positionH relativeFrom="column">
                  <wp:posOffset>-4445</wp:posOffset>
                </wp:positionH>
                <wp:positionV relativeFrom="paragraph">
                  <wp:posOffset>31750</wp:posOffset>
                </wp:positionV>
                <wp:extent cx="5495290" cy="777875"/>
                <wp:effectExtent l="0" t="0" r="0" b="0"/>
                <wp:wrapNone/>
                <wp:docPr id="1" name="Frame1"/>
                <a:graphic xmlns:a="http://schemas.openxmlformats.org/drawingml/2006/main">
                  <a:graphicData uri="http://schemas.microsoft.com/office/word/2010/wordprocessingShape">
                    <wps:wsp>
                      <wps:cNvSpPr txBox="1"/>
                      <wps:txbx>
                        <w:txbxContent>
                          <w:p>
                            <w:pPr>
                              <w:pStyle w:val="Normal"/>
                              <w:rPr/>
                            </w:pPr>
                            <w:r>
                              <w:t>Foo bar baz</w:t>
                            </w:r>
                          </w:p>
                        </w:txbxContent>
                      </wps:txbx>
                    </wps:wsp>
                  </a:graphicData>
                </a:graphic>
              </wp:anchor>
            </w:drawing>
          </mc:Choice>
          <mc:Fallback>
            <w:pict>
                <v:textbox>
                  <w:txbxContent>
                    <w:p>
                      <w:pPr>
                        <w:pStyle w:val="Normal"/>
                        <w:rPr/>
                      </w:pPr>
                      <w:r>
                        <w:t>Foo bar baz</w:t>
                      </w:r>
                    </w:p>
                  </w:txbxContent>
                </v:textbox>
              </v:rect>
            </w:pict>
          </mc:Fallback>
        </mc:AlternateContent>
      </w:r>
```
```

### DIFF
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -32,6 +32,7 @@ class PyDocXExporter(object):
 
         self.captured_runs = None
         self.complex_field_runs = []
+        self.alternate_contents = []
 
         self.node_type_to_export_func_map = {
             wordprocessing.Document: self.export_document,
@@ -69,7 +70,6 @@ class PyDocXExporter(object):
             NumberingSpan: self.export_numbering_span,
             NumberingItem: self.export_numbering_item,
             markup_compatibility.AlternateContent: self.export_markup_compatibility_alternate_content,  # noqa
-            markup_compatibility.Fallback: self.export_markup_compatibility_fallback,
             wordprocessing.TxBxContent: self.export_textbox_content,
         }
         self.field_type_to_export_func_map = {
@@ -131,6 +131,7 @@ class PyDocXExporter(object):
 
     def _post_first_pass_processing(self):
         self._convert_complex_fields_into_simple_fields()
+        self._fix_alternate_content()
 
     def _convert_complex_fields_into_simple_fields(self):
         if not self.complex_field_runs:
@@ -209,6 +210,27 @@ class PyDocXExporter(object):
             # field is now the run's new parent.
             for run in field.children:
                 run.parent = field
+
+    def _fix_alternate_content(self):
+        if not self.alternate_contents:
+            return
+        for alternate_content in self.alternate_contents:
+            new_parent_children = []
+            for child in alternate_content.parent.children:
+                if isinstance(child, markup_compatibility.AlternateContent):
+                    for alternate_content_child in alternate_content.children:
+                        # This will future-proof us in case we ever implement
+                        # markup_compatibility.Choice.
+                        child_is_fallback = isinstance(
+                            alternate_content_child,
+                            markup_compatibility.Fallback,
+                        )
+                        if not child_is_fallback:
+                            continue
+                        new_parent_children.extend(alternate_content_child.children)
+                else:
+                    new_parent_children.append(child)
+            alternate_content.parent.children = new_parent_children
 
     def export_node(self, node):
         caller = self.node_type_to_export_func_map.get(type(node))
@@ -547,7 +569,5 @@ class PyDocXExporter(object):
         return self.yield_nested(textbox_content.children, self.export_node)
 
     def export_markup_compatibility_alternate_content(self, alternate_content):
-        return self.yield_nested(alternate_content.children, self.export_node)
-
-    def export_markup_compatibility_fallback(self, fallback):
-        return self.yield_nested(fallback.children, self.export_node)
+        if self.first_pass:
+            self.alternate_contents.append(alternate_content)
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -32,7 +32,6 @@ class PyDocXExporter(object):
 
         self.captured_runs = None
         self.complex_field_runs = []
-        self.alternate_contents = []
 
         self.node_type_to_export_func_map = {
             wordprocessing.Document: self.export_document,
@@ -131,7 +130,6 @@ class PyDocXExporter(object):
 
     def _post_first_pass_processing(self):
         self._convert_complex_fields_into_simple_fields()
-        self._fix_alternate_content()
 
     def _convert_complex_fields_into_simple_fields(self):
         if not self.complex_field_runs:
@@ -210,31 +208,6 @@ class PyDocXExporter(object):
             # field is now the run's new parent.
             for run in field.children:
                 run.parent = field
-
-    def _fix_alternate_content(self):
-        if not self.alternate_contents:
-            return
-        for alternate_content in self.alternate_contents:
-            new_parent_children = []
-            for child in alternate_content.parent.children:
-                # AlternateContent has two kinds of children: Choice and
-                # Fallback. We don't care about any of the Choices. We want to
-                # replace the AlternateContent in the parent node with the
-                # content of the Fallback children.
-                if isinstance(child, markup_compatibility.AlternateContent):
-                    for alternate_content_child in alternate_content.children:
-                        # This will future-proof us in case we ever implement
-                        # markup_compatibility.Choice.
-                        child_is_fallback = isinstance(
-                            alternate_content_child,
-                            markup_compatibility.Fallback,
-                        )
-                        if not child_is_fallback:
-                            continue
-                        new_parent_children.extend(alternate_content_child.children)
-                else:
-                    new_parent_children.append(child)
-            alternate_content.parent.children = new_parent_children
 
     def export_node(self, node):
         caller = self.node_type_to_export_func_map.get(type(node))
@@ -574,4 +547,23 @@ class PyDocXExporter(object):
 
     def export_markup_compatibility_alternate_content(self, alternate_content):
         if self.first_pass:
-            self.alternate_contents.append(alternate_content)
+            new_parent_children = []
+            for child in alternate_content.parent.children:
+                # AlternateContent has two kinds of children: Choice and
+                # Fallback. We don't care about any of the Choices. We want to
+                # replace the AlternateContent in the parent node with the
+                # content of the Fallback children.
+                if isinstance(child, markup_compatibility.AlternateContent):
+                    for alternate_content_child in alternate_content.children:
+                        # This will future-proof us in case we ever implement
+                        # markup_compatibility.Choice.
+                        child_is_fallback = isinstance(
+                            alternate_content_child,
+                            markup_compatibility.Fallback,
+                        )
+                        if not child_is_fallback:
+                            continue
+                        new_parent_children.extend(alternate_content_child.children)
+                else:
+                    new_parent_children.append(child)
+            alternate_content.parent.children = new_parent_children
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -217,6 +217,10 @@ class PyDocXExporter(object):
         for alternate_content in self.alternate_contents:
             new_parent_children = []
             for child in alternate_content.parent.children:
+                # AlternateContent has two kinds of children: Choice and
+                # Fallback. We don't care about any of the Choices. We want to
+                # replace the AlternateContent in the parent node with the
+                # content of the Fallback children.
                 if isinstance(child, markup_compatibility.AlternateContent):
                     for alternate_content_child in alternate_content.children:
                         # This will future-proof us in case we ever implement
--- a/pydocx/export/base.py
+++ b/pydocx/export/base.py
@@ -567,3 +567,5 @@ class PyDocXExporter(object):
                 else:
                     new_parent_children.append(child)
             alternate_content.parent.children = new_parent_children
+            for child in new_parent_children:
+                child.parent = alternate_content.parent
--- a/pydocx/openxml/markup_compatibility/fallback.py
+++ b/pydocx/openxml/markup_compatibility/fallback.py
@@ -12,4 +12,8 @@ from pydocx.openxml.wordprocessing.picture import Picture
 class Fallback(XmlModel):
     XML_TAG = 'Fallback'
     # TODO #204: actually include all of the children defined in the spec.
-    children = XmlCollection(Picture)
+    children = XmlCollection(
+        'wordprocessing.Run:children',
+        'wordprocessing.Body:children',
+        'wordprocessing.Run',
+    )
--- a/pydocx/openxml/markup_compatibility/fallback.py
+++ b/pydocx/openxml/markup_compatibility/fallback.py
@@ -13,5 +13,4 @@ class Fallback(XmlModel):
     children = XmlCollection(
         'wordprocessing.Run:children',
         'wordprocessing.Body:children',
-        'wordprocessing.Run',
     )
--- a/pydocx/openxml/markup_compatibility/fallback.py
+++ b/pydocx/openxml/markup_compatibility/fallback.py
@@ -10,4 +10,9 @@ from pydocx.models import XmlModel, XmlCollection
 
 class Fallback(XmlModel):
     XML_TAG = 'Fallback'
+
+    # It would be better to refer to the grandparent's children XmlCollection
+    # lazily. The problem is that we don't have a good way to represent lazy
+    # fields, nor do we have a way for those fields to reference their parents
+    # before they are initialized with content.
     children = XmlCollection(allow_all_children=True)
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by kylegibson** [comment]
Could you add a body-level fallback test case?

**Comment by caffodian** [comment]
👍

**Comment by jlward** [comment]
> Could you add a body-level fallback test case?

What do you mean?

File: tests/export/html/test_markup_compatibility.py [hunk 1]
```diff
@@ -91,3 +91,71 @@ def test_textbox_with_content_outside_of_textbox(self):
...
+
+    def test_fallback_is_only_text(self):
+        document_xml = '''
+            <p>
+                <r><t>AAA</t></r>
+                <r>
+                    <t>BBB</t>
+                    <AlternateContent>
+                        <Fallback>
+                            <r>
```
kylegibson: My understanding was that the fallback would need to only contain those nodes that could otherwise be contained by the grandparent. In this case, you have a fallback with a run. The grandparent is a run. So you have nested runs, which isn't valid. In this case, assuming my understanding is correct (which it may not be), what you really want to test is `<Fallback><t>CCC</t></Fallback>`.

tests/export/html/test_markup_compatibility.py
```diff
@@ -91,3 +91,71 @@ def test_textbox_with_content_outside_of_textbox(self):
...
+
+    def test_fallback_is_a_table(self):
+        document_xml = '''
+            <p>
+                <r><t>AAA</t></r>
+                <r>
+                    <t>BBB</t>
+                    <AlternateContent>
+                        <Fallback>
+                            <tbl>
```
kylegibson: I don't understand how this would be supported either, since tbls are children of the body, not runs. Maybe I'm missing something here?

jlward: I haven't been able to find any documentation for where `AlternateContent` can or cannot be. So I am kind of winging it.

kylegibson:
> What do you mean?

I am under the impression that AlternativeContent can appear anywhere within the document structure. Is that true or false? 

I'm suggesting that you have a test case in which the fallback includes a paragraph with some text, for example:

So it might look like this:

```
Body
   p, r, t, foo
   AlternateContent
      Choice
          ...
      Fallback
          p, r, t, bar
    p, r, t, baz
```

Does that make sense? I also think it's a good idea to have one or more tests that include the Choice too.

tests/export/html/test_markup_compatibility.py
jlward: As such, I'm assuming it can have the same children as a run and the body tag. Because those should cover most cases.

jlward:
> I also think it's a good idea to have one or more tests that include the Choice too.

We are blatantly ignoring the `Choice` so I don't see any reason to include it.

tests/export/html/test_markup_compatibility.py
kylegibson: I am thinking that AC can existing anywhere in the document structure. The fallback inherits it's grandparents children.

kylegibson:
> so I don't see any reason to include it.

To test that we are ignoring it...

jlward:
> To test that we are ignoring it...

I added a test, if we ever implement `Choice`, which we theoretically could, this test will not be guaranteed to pass/fail as expected. Since the choice is depenedent on

tests/export/html/test_markup_compatibility.py
jlward: Building grandparents is 10x harder than picking and choosing possible nodes. Since it's 100% broken right now, this solution is better than nothing.

kylegibson:
[Ecma Office Open XML Part 3 - Markup Compatibility and Extensibility.pdf](https://github.com/CenterForOpenScience/pydocx/files/266948/Ecma.Office.Open.XML.Part.3.-.Markup.Compatibility.and.Extensibility.pdf)

AlternateContent and Fallback are documented in section 10.2

kylegibson:
From the reference above:

> The content of the selected Choice or Fallback element is processed as though it replaces the entire
AlternateContent element.

jlward:
> From the reference above:

> > The content of the selected Choice or Fallback element is processed as though it replaces the entire
AlternateContent element.

We don't really have an easy way to do this. Which means this issue will almost certainly be punted.

tests/export/html/test_markup_compatibility.py
kylegibson: Given that it's been 100% broken for so long, I don't think we're under any rush to just do the bare minimum to get this working. I'm not convinced this would be 10x harder.

jlward: To inherit the grandparents children requires that we know what the grandchild is. As such, it cannot be defined at the class level. If we are in no rush to fix this, then I am going to punt on this.

jlward: The easiest way to do this would be to destroy the choice element and un wrap the alternative content and the fallback content. I see two problems with that:

1. AlternateContent would basically need to be a child of all elements in order to work correctly
1. I have no idea how to alter the XML structure. I see that we are doing it for hyperlinks, I have no idea how it works.

kylegibson: > To inherit the grandparents children requires that we know what the grandchild is

It requires that the node know what it's grandparent is, which you are correct, cannot be done at the class level without some kind of special syntax. 

An additional problem is that the AlternateContent can exist as a child of any node. Currently we have it in (at least) the run node. It's not practical to implicitly add it to every node, and I'm not sure how we would ensure that it was added in every new node. 

> If we are in no rush to fix this, then I am going to punt on this.

While there's no rush, if we upgrade libreoffice without fixing this, it will mean some percentage of documents will be busted. That percentage may be low enough that it doesn't matter. You'll need to run through the QA test to see what the impact is.

kylegibson: > AlternateContent would basically need to be a child of all elements in order to work correctly

Yes, that is correct. Based on my understanding of the reference, that is how it should be.

jlward: Forcing `AlternateContent` to be a valid child of things would actually not be that difficult. We have the framework in place to do that dynamically.

jlward: Which leaves the actual hard part of: I don't know how to alter the XML. Again, I know we do it, I don't understand how it works.

kylegibson: Is there a specific part that you do not understand?

jlward: https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/export/base.py#L135

I don't understand how that method makes changes in a way that get reflected in the second pass.

kylegibson: OK, I see. 

The key changes are https://github.com/CenterForOpenScience/pydocx/blob/master/pydocx/export/base.py#L202 and the lines below that. Everything else before that point is (mostly) just building and mapping the necessary objects.

It might help you if you insert a set_trace and step through the function using one of the existing complex field test cases.

jlward: Alright, I'll see what I can do. Thanks for the help :)

jlward: Alright, I have most of this actually figured out. The only pickle is the valid children of the `Fallback` element. Which is actually odd now that I think about it. Since we are should be removing the `Fallback` element and the re-parse should do everything we need. So perhaps I'm missing something? Or I did it wrong? Not sure.

kylegibson: Nice :) That was a quick 19 minutes :)

I'm thinking we remove the node in the first pass.

jlward: > I'm thinking we remove the node in the first pass. 

What do you mean?

kylegibson: The code you've added doesn't appear to actually fully remove the node. You're not updating the parent relationships.

jlward: > The code you've added doesn't appear to actually fully remove the node. You're not updating the parent relationships.

And you think that will result in us not caring about the children of the `Fallback` element? Also, I'm editing the parents children. I assume I need to do something like set the parent to None for the `AlternateContent` and `Fallback`? Although I can't really see how that would help

kylegibson: I'm not sure how it would need to be implemented without spending more time on it. But you'll need to do some management of the parent relationships somewhere.

pydocx/openxml/markup_compatibility/fallback.py
```diff
@@ -6,10 +6,8 @@
...
 
 from pydocx.models import XmlModel, XmlCollection
-from pydocx.openxml.wordprocessing.picture import Picture
 
 
 class Fallback(XmlModel):
     XML_TAG = 'Fallback'
-    # TODO #204: actually include all of the children defined in the spec.
-    children = XmlCollection(Picture)
+    children = XmlCollection(allow_all_children=True)
```
kylegibson: Here's another idea: Make `children` a property. In the property, goes up two parent levels, and returns the children from that level. Would that work?

pydocx/export/base.py
```diff
@@ -547,7 +573,5 @@ def export_textbox_content(self, textbox_content):
         return self.yield_nested(textbox_content.children, self.export_node)
 
     def export_markup_compatibility_alternate_content(self, alternate_content):
-        return self.yield_nested(alternate_content.children, self.export_node)
-
-    def export_markup_compatibility_fallback(self, fallback):
-        return self.yield_nested(fallback.children, self.export_node)
+        if self.first_pass:
+            self.alternate_contents.append(alternate_content)
```
kylegibson: I don't quite understand why alternate contents need to be captured. We had to capture complex fields because they would span across runs. That doesn't seem to be the case with ACs. I was expecting you would have a method here like `self._unwrap_alternate_content_fallback` and you would pass in the alternate content.

Then, that method would update the parent of each child in the fallback to match the AC's parent, and add them as children of the AC's parent.

pydocx/openxml/markup_compatibility/fallback.py
jlward: self.parent.parent.chlidren is a list. Not an `XmlCollection`. So no, I don't think that will work.

pydocx/export/base.py
jlward: I don't understand the difference between what you suggested and what I have.

pydocx/openxml/markup_compatibility/fallback.py
kylegibson: You get the parents model class and return the children field, not the instance

pydocx/export/base.py
kylegibson: My suggestion doesn't involve capturing all of the AC's. Instead, I am suggesting you process each AC individually, here. Does that make sense?

kylegibson: So you wouldn't be using `self.alternate_contents`

pydocx/openxml/markup_compatibility/fallback.py
jlward: Still doens't work. All the content from the `Fallback` is gone.

pydocx/export/base.py
jlward: I'd rather not edit the structure of the XML while we are parsing it. I think it makes more sense to do the altering in between passes.

pydocx/openxml/markup_compatibility/fallback.py
kylegibson: Without seeing how you implemented it, I can only guess at what might be wrong.

jlward:
```python
class Fallback(XmlModel):
    XML_TAG = 'Fallback'

    # children = XmlCollection(allow_all_children=True)

    @property
    def children(self):
        print(self.parent.parent, dir(self.parent.parent))
        return self.parent.parent.__class__.children.types
```

kylegibson: What happens if you just return `self.parent.parent.__class__.children` ?

jlward:
```
  File "/home/jason/third_party_apps/pydocx/pydocx/test/testcases.py", line 235, in run_test
    result = self.convert_docx_to_html(docx_path)
  File "/home/jason/third_party_apps/pydocx/pydocx/test/testcases.py", line 250, in convert_docx_to_html
    return exporter.export()
  File "/home/jason/third_party_apps/pydocx/pydocx/export/html.py", line 211, in export
  File "/home/jason/third_party_apps/pydocx/pydocx/export/html.py", line 209, in <genexpr>
  File "/home/jason/third_party_apps/pydocx/pydocx/export/base.py", line 119, in export
    self._post_first_pass_processing()
  File "/home/jason/third_party_apps/pydocx/pydocx/export/base.py", line 134, in _post_first_pass_processing
    self._fix_alternate_content()
  File "/home/jason/third_party_apps/pydocx/pydocx/export/base.py", line 234, in _fix_alternate_content
    new_parent_children.extend(alternate_content_child.children)
TypeError: 'XmlCollection' object is not iterable
```

kylegibson: Okay, that's good and bad news. Good news is that the property works. Bad news is that the model processor isn't properly replacing the field with instances. I don't know why that is. I would need to put a set_trace in the XmlModel.load function to see what's happening to it.

kylegibson: The benefit of this approach is that if it works, you won't need the logic for allowing all types as children.

jlward: Sounds a lot like I should add a TODO comment with this better solution, and move on. Thoughts?

kylegibson: I personally would spend at least an hour trying to figure out why it's not evaluating the field correctly before punting on it. The cost of merging in complicated code that may get used by others before we fix it is pretty high.

pydocx/export/base.py
winhamwr: Kyle is suggesting altering our data model of the XML, not the XML itself, right? That's definitely a cleaner solution than adding more global state (self.alternate_contents) that triggers only on the first pass.

pydocx/openxml/markup_compatibility/fallback.py
jlward: Spent the last hour and a half trying to figure this out. No luck.

kylegibson: OK. I just also spent a half hour. I see what we need to do, but it's somewhat complicated. Can you add a comment here explaining that it would be better to refer to the grandparent's `children` XmlCollection lazily? The problem is that we don't have a good way to represent lazy fields, nor do we have a way for those fields to reference their parents before they are initialized with content.

jlward: I thought I had already added a comment, I meant to. Give me a moment.

tests/export/html/test_markup_compatibility.py
kylegibson: tables aren't valid run children

kylegibson:
That we're not currently able to build Fallback to spec bothers me a bit. Oh well. This looks good otherwise. Thanks for your patience on this issue. I'll create another open issue describing what we need to do to make Fallback work to spec.

PR #216, opened 2016-05-20
Picts can have rects, which can have textbox

jlward:
```
<mc:AlternateContent>
  <mc:Fallback>
    <w:pict>
      <v:rect fillcolor="#FFFFFF" strokecolor="#000000" strokeweight="1pt" style="position:absolute;rotation:0">
        <v:textbox>
          <w:txbxContent>
            <w:p>
              <w:r>
                <w:t>AAA</w:t>
              </w:r>
            </w:p>
          <w:txbxContent>
        <v:textbox>
      <v:rect>
    <w:pict>
  <mc:Fallback>
</mc:AlternateContent>

```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,7 @@
+**0.9.8**
+
+- Pict elements can now contain Rect elements.
+
 **0.9.7**
 
 - Text colors other than black and white are no longer ignored
--- a/pydocx/__init__.py
+++ b/pydocx/__init__.py
@@ -6,4 +6,4 @@ __all__ = [
     'PyDocX',
 ]
 
-__version__ = '0.9.7'
+__version__ = '0.9.8'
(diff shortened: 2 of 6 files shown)


pydocx/openxml/wordprocessing/picture.py
```diff
@@ -6,10 +6,10 @@
...
 from pydocx.models import XmlModel, XmlCollection
-from pydocx.openxml.vml import Shape
+from pydocx.openxml.vml import Shape, Rect
 
 
 class Picture(XmlModel):
     XML_TAG = 'pict'
 
-    children = XmlCollection(Shape)
+    children = XmlCollection(Shape, Rect)
```
caffodian: So, as a noob trying to understand, we're looking at something like this https://msdn.microsoft.com/en-us/library/documentformat.openxml.wordprocessing.picture(v=office.15).aspx which states all the various VML shapes can sit under it?

caffodian:
👍

pydocx/openxml/wordprocessing/picture.py
jlward: That is correct. Some we don't care about, some we haven't built yet.

Issue #221, opened 2016-08-25
Internal links are simply dropped


sunu:
Hi, first of all thanks a lot for making this library. Really appreciate the effort you folks have put in.

I'm trying to convert a docx file with internal links and I notice all the links are simply dropped in the html output. I have attached the simplest test case to reproduce this.
[test.docx](https://github.com/CenterForOpenScience/pydocx/files/436346/test.docx)

I did a bit of investigation on my own and it seems at https://github.com/CenterForOpenScience/pydocx/blob/3d01b3c7210f78a7c69409feac8c2cc8663d7de1/pydocx/export/html.py#L510 the value of tag is always `None` for internal links because `hyperlink.target_uri` seems to be always `None`. The only unique value identifying an internal link seems to be stored in `hyperlink.anchor`, but I'm not sure how to identify the target from that value.

I would love to see a fix for this. If you folks can give me some pointers on how to fix this, I'm all in for writing up a PR for this if you like.

Thanks.

winhamwr:
Fixed with #222

PR #222, opened 2016-08-25
Don't drop internal links

sunu:
I tried fixing #221 and this is the rough solution I have for now. If this looks ok, I can polish it up and add tests.

### DIFF
--- a/pydocx/export/html.py
+++ b/pydocx/export/html.py
@@ -270,6 +270,11 @@ class PyDocXHTMLExporter(PyDocXExporter):
             heading_style.name.lower(),
             self.default_heading_level,
         )
+        if paragraph.bookmark_name:
+            attrs = {
+                'id': paragraph.bookmark_name
+            }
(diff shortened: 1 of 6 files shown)


pydocx/export/html.py
```diff
@@ -270,6 +270,11 @@ def get_heading_tag(self, paragraph):
             heading_style.name.lower(),
             self.default_heading_level,
         )
+        if paragraph.bookmark_name:
+            attrs = {
+                'id': paragraph.bookmark_name
+            }
+            return HtmlTag(tag, **attrs)
```
winhamwr: How about:
`return HtmlTag(tag, id=paragraph.bookmark_name`

winhamwr:
Hello Tarashish,

This looks very reasonable to me. Add some tests and then I think this is ready for merging.

-Wes

sunu:
@winhamwr Hey! Thanks for the quick feedback. I have added tests now.

Issue #224, opened 2016-12-21
Odd margin being added to bullets


jhubert:
When certain docx files that have adjusted margins get imported, the resulting HTML places the margin in the wrong place. This results in oddly formatted HTML.

For example, here are two lists in word:

![image](https://cloud.githubusercontent.com/assets/3384/21377620/8e32446e-c705-11e6-8a96-58e63e7ddc38.png)

The first list has been indented, the second one has the standard doc indentation.

Here is the result in HTML:

![image](https://cloud.githubusercontent.com/assets/3384/21377637/acb68422-c705-11e6-9987-3a371958eac0.png)

The resulting HTML has a span inside the `li` with a `margin-left` set on it:

```html
<li><span style="margin-left:3.00em">This is a list item</span></li>
```

It seems that the whole `ul` should have the margin, if anything at all.

Here is the sample file:
[list-item-margin.docx](https://github.com/CenterForOpenScience/pydocx/files/665491/list-item-margin.docx)

And here is the cleaned up docx source from the `document.xml` file:

```xml
  <w:p w14:paraId="25B98899" w14:textId="77777777" w:rsidR="00442583" w:rsidRDefault="00442583" w:rsidP="00442583">
    <w:r>
      <w:t>Headline:</w:t>
    </w:r>
  </w:p>
  <w:p w14:paraId="59BA247B" w14:textId="77777777" w:rsidR="00442583" w:rsidRDefault="00442583" w:rsidP="00442583">
    <w:pPr>
      <w:pStyle w:val="ListParagraph"/>
      <w:numPr>
        <w:ilvl w:val="0"/>
        <w:numId w:val="2"/>
      </w:numPr>
      <w:ind w:left="720"/>
    </w:pPr>
    <w:r>
      <w:t>This is a list item</w:t>
    </w:r>
  </w:p>
  <w:p w14:paraId="551666C9" w14:textId="77777777" w:rsidR="00442583" w:rsidRDefault="00442583" w:rsidP="00442583">
    <w:pPr>
      <w:pStyle w:val="ListParagraph"/>
      <w:numPr>
        <w:ilvl w:val="0"/>
        <w:numId w:val="2"/>
      </w:numPr>
      <w:ind w:left="720"/>
    </w:pPr>
    <w:r>
      <w:t>This is a list item</w:t>
    </w:r>
  </w:p>
[...]

Issue #202, opened 2016-02-01
Format section/subsection - missing tabs/spaces.


botzill:
Hi.

I have such a file:

[subsections_format.docx](https://github.com/CenterForOpenScience/pydocx/files/112134/subsections_format.docx)

after converting to html we get:

![2ea64158-ae83-11e5-8ed4-d56bff32bef1](https://cloud.githubusercontent.com/assets/4686040/12716344/6d35b8d4-c8e8-11e5-9edb-a32d3088b7e5.png)

As you can see the subsections are not properly formatted. If you guide me where to look for this issue I can submit a pull request to solve this. 

Thx a lot.

jhubert:
## Before:

![image](https://cloud.githubusercontent.com/assets/3384/13727369/c5698a32-e8a3-11e5-87cf-993d25b0d6ff.png)


## After:

![image](https://cloud.githubusercontent.com/assets/3384/13727370/da78e2ec-e8a3-11e5-8212-368d24ada99f.png)

winhamwr:
It looks like there is one definite bug, along with some confusion about the styling.

"Gather Items for Re-pricing" should definitely be in the same list as "Prepare your markdown gun" and it's not obvious to me why it isn't.

The first step will be adding a [fixtures testcase](https://github.com/CenterForOpenScience/pydocx/blob/master/tests/export/test_docx.py#L25) by adding both a .docx and .html file in the [fixtures](https://github.com/CenterForOpenScience/pydocx/tree/master/tests/fixtures) directory. That will let us define the input and then the expected output.

If anyone could help with that part, it would be appreciated. From there, someone will need to dive in to the OOXML in the .docx to figure out why we're parsing the .docx as separate lists instead of one list.

jhubert:
I dove in and took a look at the OOXML for this. I've added the fixtures as well.

It looks like what's happening is that it's being considered three different lists because the bulleted list is breaking up the numeric list.

![image](https://cloud.githubusercontent.com/assets/3384/13793728/9724c7c2-eab5-11e5-8010-3d9e987170c0.png)

Here is the _simplified_ relevant `document.xml` OOXML:

```xml
<w:p><w:pPr><w:numPr><w:ilvl w:val="0"/><w:numId w:val="1"/></w:numPr></w:pPr><w:r><w:t>one</w:t></w:r></w:p>
[...]

PR #225, opened 2017-02-01
Fixed item listing, moved margin-left to <li>

botzill:
No unit tests yet! Will add soon.

About this PR. Basically there are scenarios like in the issue #224 and #202. 

Scenario 1 - proper left/right margin:

Currently when we have a doc like:

screen shot 2017-02-01 at 7 33 08 pm

(old code) will output this:

screen shot 2017-02-01 at 7 33 59 pm

(new code) will output this:

screen shot 2017-02-01 at 7 35 54 pm

So, basically proper margins are added to the `<li>` items like this:

```
<body><p>Heading 1</p>
<ol class="pydocx-list-style-type-decimal">
    <li style="margin-left:9.00em">AAA</li>
    <li style="margin-left:9.00em">BBB</li>
</ol>
<p>Heading 2</p>
<ul>
    <li style="margin-left:19.50em">CCCC</li>
    <li style="margin-left:19.50em">DDD</li>
</ul>
</body>
```

Scenario 2 - basically issue from #202:

If we have as input a list like this:

screen shot 2017-02-01 at 7 43 48 pm

(old code) we get as output this:

screen shot 2017-02-01 at 7 44 58 pm

(new code) while with new changes we get:

screen shot 2017-02-01 at 7 45 34 pm

Now, this issues is related to the fact that listings like: `B1,B2`, `C1,C2` are basically independent lists from the main one. This mean that they have a different ``NumberingProperties`` and, of course, different ``numId`` which is actually breaking the entire listing flow. Currently listing will works if ``NumberingProperties`` are the same and ``level_id`` is changing(which will cause proper nesting). Now, for this case I did the following(code commends as well https://github.com/CenterForOpenScience/pydocx/pull/225/commits/5a4d93a2d8f11553731119781fe1985b322f9e95#diff-4aa70f6d5310ea0a57e5c20b629dc727R212):

Before we start to process numbering spans/items I read all the listing items(from ``components``) and try to determine such scenarios and create a simple mapping like this:

```
child_parent_num_map = {
     "<child_num_id>": "<parent_num_id>",
    ....
}
parent_child_num_map = {
     "<parent_num_id>": "<child_num_id>",
    ....
}
```  

Later, when we construct the span/items I do some checks and determine if we should start a new span or we already have a parent one and just add to it.
Check PR for more details relate to implementation and let me know what you think about this solution.

 

 
Now, this issues was tricky as I needed to understand basically the entire listing flow. Still not sure if I did it properly


winhamwr:
Hello Chirica,

Thanks for the PR! 

There's a lot going on here. It would really help if you could provide both a summary of what was wrong and what the new behavior is, and at least some minimal test coverage demonstrating the fix. 

Thanks
-Wes

kylegibson:
@winhamwr I think the discussion is in #224

botzill:
Ou, I did not expect the PR here :). OK, I will make proper descriptions about it. Still some work in progress here. So, hold on for now. 

Thx.

winhamwr:
Thanks @kylegibson! The discussion there is great.

I guess that means we're just missing a bit of test coverage and a thorough review. I walked through the code itself and it looks high-quality. Thanks for that, Chirica! 

> So, hold on for now.

Will-do. We'll wait to hear from you before diving in.

botzill:
OK, I did update the description with that I did so far. Please, if you have time, check the PR and let me know if you have any improvements. I did not covered 100% how listings are get constructed(fake lists as example). 

Thx!

botzill:
@winhamwr @kylegibson any chance you can check the current PR an let me know what you think about it? 

Thx!

tests/export/test_xml.py
```diff
@@ -461,27 +461,29 @@ def get_xml(self):
...
-        <ol class="pydocx-list-style-type-lowerLetter">
-            <li>CCC</li>
+            <li>
+                AAA
+                <ol class="pydocx-list-style-type-decimal">
+                    <li>BBB</li>
+                    <li>CCC</li>
+                </ol>
+            </li>
+            <li>DDD</li>
```
winhamwr: This testcase really helped me understand the new behavior and how it's superior to the previous behavior. Thanks for including it!

pydocx/export/html.py
```diff
@@ -311,46 +311,46 @@ def export_paragraph_property_justification(self, paragraph, results):
...
-                first_line = convert_twips_to_ems(first_line)
-                # TODO text-indent doesn't work with inline elements like span
-                style['text-indent'] = '{0:.2f}em'.format(first_line)
+        if indentation_left:
+            left = convert_twips_to_ems(indentation_left)
+            style['margin-left'] = '{0:.2f}em'.format(left)
+
+        if indentation_first_line:
+            first_line = convert_twips_to_ems(indentation_first_line)
+            # TODO text-indent doesn't work with inline elements like span
```
winhamwr: Can we remove this TODO, now that we're adding `display: inline-block`?

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+            if level_id == 0:
+                prev_level_id = level_id
+            else:
+                prev_level_id = level_id - 1
+
+            prev_level_paragraphs = self.numbering_level_listing_track[num_id][prev_level_id]
+            if prev_level_paragraphs:
+                return prev_level_paragraphs[-1]
+
+            if prev_level_id == 0 and not prev_level_paragraphs:
```
winhamwr: Do we need the `and not prev_level_paragraphs` condition? Seems like we'd get the same behavior without it.

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+                # this is an edge case with older version of word when it may contain a sublist
```
winhamwr: Thanks for the "why" comment! Can we capitalize the `t` in `This` to give it the treatment of a proper english sentence?

pydocx/export/html.py
```diff
@@ -311,46 +311,46 @@ def export_paragraph_property_justification(self, paragraph, results):
...
 
-        if properties.indentation_right:
-            # TODO would be nice if this integer conversion was handled
-            # implicitly by the model somehow
-            try:
-                right = int(properties.indentation_right)
-            except ValueError:
-                right = None
+        # for numbering properties we add style to span item level
+        if properties.numbering_properties is None:
```
winhamwr: How would you feel about switching this conditional so that it was `if properties.numbering_properties` so that the `else` case was the 2-line case? Just a matter of preference, but I struggled for a bit reading this until I realized that the first case was essentially the "do nothing fancy" case.

I'd also love a comment explaining the do-something case like `Numbering properties can define a text indentation on a paragraph`

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+            if prev_level_id == 0 and not prev_level_paragraphs:
+                # this is an edge case with older version of word when it may contain a sublist
+                # into a separate num_id.
+                break
+
+            level_id -= 1
+
+        return None
+
+    def export_listing_paragraph_property_indentation(self, paragraph, level_properties,
```
winhamwr: The (undocumented, sorry) coding standard this project follows is to use one line per kwarg when the definition stretches across lines.

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+        paragraph_ind_left = paragraph_properties.to_int('indentation_left', default=0)
+        paragraph_ind_hanging = paragraph_properties.to_int('indentation_hanging', default=0)
+        paragraph_ind_first_line = paragraph_properties.to_int('indentation_first_line',
+                                                               default=0)
+
+        left = paragraph_ind_left or level_ind_left
+        hanging = paragraph_ind_hanging or level_ind_hanging
+
+        # at this point we have no info about indentation, so we keep the default one
+        if not left and not hanging:
```
winhamwr: Are we ignoring the possibility that only `paragraph_ind_first_line` is set on purpose? Is that case handled with [this line](https://github.com/CenterForOpenScience/pydocx/pull/225/files#diff-1e8de747886c45532983f883b63b0b84R322)?

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+            # by default there are only 9 numbering levels in docx(0 indexed)
+            self.numbering_level_listing_track[num_id] = [[] for _ in range(10)]
+        if paragraph not in self.numbering_level_listing_track[num_id][level_id]:
+            self.numbering_level_listing_track[num_id][level_id].append(paragraph)
+
+        # by default left contains hanging as well, so we remove it
+        left -= hanging
+
+        if level_id == 0:
+            # because html ul/ol/li elements have there default indentations
```
winhamwr: there => their

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+            # we remove the default word one as well
```
winhamwr: Minor sentence cleanup:
We remove...
...as well.
This way...
...to html.

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+
+            # first line are added left margins
+            if paragraph_ind_first_line:
+                left += paragraph_ind_first_line
+
+        if level_id > 0:
+            # for nested levels we need to add indentation based on parent level
+            prev_paragraph = self.get_previous_level_paragraph(num_id, level_id)
+            if prev_paragraph:
+                prev_left_level_indentation = prev_paragraph.get_numbering_level().\
```
winhamwr: We go with the style of opening the parens and then indenting the next line, rather than using `\`.

pydocx/export/html.py
```diff
@@ -361,6 +361,114 @@ def export_paragraph_property_indentation(self, paragraph, results):
...
+
+        # we don't allow negative hanging
+        if hanging < 0:
+            hanging = 0
+
+        if include_text_indent:
+            if hanging is not None:
+                # Now, here we add the hanging as text-indent for the paragraph
+                hanging = convert_twips_to_ems(hanging)
+                # TODO text-indent doesn't work with inline elements like span
```
winhamwr: Do we want to add the `display: inline-block` bit here, too?

pydocx/export/numbering_span.py
```diff
@@ -206,6 +208,103 @@ def get_numbering_level(self, paragraph):
             return None
         return level
 
+    def detect_parent_child_map_for_items(self):
```
winhamwr: @jlward @kylegibson would either of you be willing to give this method a look? I understand what this is supposed to do (thanks to the excellent docstring), and I understand the code (looks good), but I don't understand the interaction between numbering_properties, abstract_num_id, and other concepts here to give it a confident review.

pydocx/export/html.py
botzill: It is on a new line no?

winhamwr: Sorry that I wasn't clear. Like this:

```python
def export_listing_paragraph_property_indentation(
    self, 
    paragraph, 
    level_properties, 
    include_text_indent=False,
):
```

Actually, don't worry about these minor style issues, including my notes on comment capitalization/punctuation. I can handle all of that trivial stuff after we get this merged in. We can't expect you to read our mind on those things, since they aren't documented in our [coding standards](https://pydocx.readthedocs.io/en/latest/development.html#coding-standards).

pydocx/export/html.py
botzill: Well, here we add margins to `li` elements, so no need at this stage.

pydocx/export/html.py
botzill: like this:

```python
prev_left_level_indentation = prev_paragraph.get_numbering_level(
    ).paragraph_properties.to_int('indentation_left')
```
?

pydocx/export/html.py
botzill: Well, if paragraph is part of the listing then it's level will contain the `left` or `hanging`. I don't think we can have a listing paragraph where we don't have `left/hanging` on it's level and on paragraph but first line is set.

pydocx/export/html.py
botzill: Nop, good to know ;).

pydocx/export/numbering_span.py
botzill: I just realized that this methods needs to be adapted to handle more cases, especially any nested separated list. Will look into it tomorrow and try to adapt for all cases. Any feedback is welcome.

winhamwr:
We've given this a couple of days for review, and everything looks very solid to me. @botzill if you can get the python 2.6 tests passing, then I'll merge this in and fix the minor style issues.

botzill:
Thx @winhamwr, sounds great! I have left some final things to finish related to setting proper margins for those separated nested lists. Hope will not take much time. I want to include that as well in this PR. 

p.s Tests fixed.

botzill:
OK, I think this PR is ready now. Did some changes related to separated nested lists. Now margins are properly set.

PR #231, opened 2017-02-09
Add table cell background color

botzill:


botzill:
@winhamwr Any chance you can take a look at this PR as well ?

pydocx/openxml/wordprocessing/table_cell_properties.py
```diff
@@ -25,3 +27,13 @@ def should_close_previous_vertical_merge(self):
         if merge != 'continue':
             return True
         return False
+
+    @property
+    def background_color(self):
+        background_fill = getattr(self, 'background_fill', None)
+
+        # There is no need to set white background color
+        if background_fill not in ('auto', 'FFFFFF'):
```
jhubert: Does it make sense to reuse COLOR_FOR_DARK_BACKGROUND here?

jhubert: Does it make sense to reuse COLOR_FOR_DARK_BACKGROUND here?

botzill: I guess we can use, word color is also in RRGGBB format, good catch 👍 .

